PR #2, opened 2016-10-25
Added Vagrantfile & provisioning

davesims:
#### Vagrantfile & Provisioning Script 

This PR will add the Vagrantfile to be used as the basis for a kubos development environment. The base image will be built by Jenkins using the provisioning script (`provision.sh`), and published to Vagrant's Atlas under our kubostech account. 

The Responsibility of the Vagrantfile will be to add the ability for Virtualbox to mount USB devices so that kubos/yotta can flash to the attached microcontroller.

TODO:
- [x] install python, pip, py dev tools, git
- [x] install compilation/debug tools
- [x] git clone kubos (this repo) to `~/.kubos`
- [x] Install the CLI
- [x] put `kubos` on the PATH

/cc @kubostech/devs @kyleparrott @plauche


davesims:
@kyleparrott @plauche Can you point me to the compilation binaries we'll need?

kyleparrott:
@davesims: A good reference for our current config is our [Dockerfiles for the kubos-sdk images](https://github.com/kubostech/KubOS-legacy/blob/master/tools/images/Dockerfile.build)

That list won't include the openocd, dfu-util or mspdebug packages which we will need as well.

And [libmsp430.so](https://github.com/kubostech/kubos-sdk/tree/master/lib/linux)

plauche:
Do we want to store these vm related files in this repo? Currently this repo is only storing source modules.

davesims:
@plauche We could put it in the cli, or in its own repo. @kyleparrott and I talked about that yesterday, didn't really come to a solid answer. I think here may make the most sense, since changes to the targets will need sometimes to be reflected in the compilation environment. It seems to me the provisioning script & source modules are coupled together in that way.

davesims:
@plauche @kyleparrott I'm going to leave it here for now, we can move it later if we find a better scheme.

davesims:
@kyleparrott @plauche Still very much WIP, but can you see anything else needed to get the compilation environments up to speed?

plauche:
@davesims You will also need to install `openocd`, `dfu-util`, and `mspdebug`. The first two can be installed via `apt`. The mspdebug is most reliably borrow from the [Energia Project](http://energia.nu/download/). The binary can be found [here - mspdebug](https://github.com/kubostech/kubos-sdk/tree/master/bin/linux), and the needed library [here - libmsp430.so](https://github.com/kubostech/kubos-sdk/tree/master/lib/linux).

vm/Vagrantfile
```diff
@@ -0,0 +1,39 @@
...
+      apt-get -y install python-pip
+      add-apt-repository ppa:git-core/ppa
+      apt-get -y install git
+      add-apt-repository ppa:team-gcc-arm-embedded/ppa
+      add-apt-repository ppa:george-edison55/cmake-3.x
+      apt-get install -y cmake=3.2.2-2~ubuntu14.04.1~ppa1
+      apt-get install -y gcc-msp430 gdb-msp430 msp430-libc
+
+      mkdir -p /home/vagrant/.kubos
+      easy_install pip
```
kyleparrott: We're install pip twice. The python-pip package is installed on line 15 and then we install pip using easy_install here. We should check to see which version of pip these methods install and should update pip if necessary. `pip install --upgrade pip`

test/integration/test_cli.py
```diff
@@ -0,0 +1,16 @@
+import unittest
+
+class TestCLI(unittest.TestCase):
+
+    def test_kubos_installed(self):
+        self.assertEqual('foo'.upper(), 'FOO')
```
plauche: I'm guessing these are just place holder test statements?

davesims: Yep, new test code just pushed. WIP ftw!

PR #5, opened 2016-10-26
KUBOS-187 Adding Debug Configuration for the stm32f407 discovery target

kyleparrott:
Very basic debugging implementation for the stm32f4 discovery board.

I'm not crazy about the debugging implementation, specifically how openocd is started and not managed very well, or really at all by this implementation. It does work but should definitely be improved soon.

### DIFF
--- a/targets/target-stm32f407-disco-gcc/openocd/flash.sh
+++ b/targets/target-stm32f407-disco-gcc/openocd/flash.sh
@@ -1,27 +1,40 @@
 #!/bin/bash
 
 this_dir=$(cd "`dirname "$0"`"; pwd)
-cmd=$1
+program=$1 # program is fist becasue for debugging there is not a command provided
+           # openocd will start a gdb server by default if there's not a command provided
+cmd=$2
+if [[ ! -z $cmd ]]; then
+    openocd_arg="$cmd $program"
+fi
 unamestr=`uname`
 
 if [[ "$unamestr" =~ "Linux" ]]; then
-	device=`lsusb -d '0483:'`
+    device=`lsusb -d '0483:'`
 elif [[ "$unamestr" =~ "Darwin" ]]; then
-	device=`python $this_dir/osxusb.py -d '0483:'`
+    device=`python $this_dir/osxusb.py -d '0483:'`
 fi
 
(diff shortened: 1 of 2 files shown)


kyleparrott:
In the future I want to transition this to some sort of python script that uses [pyOCD](https://pypi.python.org/pypi/pyOCD/0.5.1) or something similar to manage the openOCD gdb server.

targets/target-stm32f407-disco-gcc/openocd/flash.sh
```diff
@@ -1,27 +1,40 @@
 #!/bin/bash
 
 this_dir=$(cd "`dirname "$0"`"; pwd)
-cmd=$1
+program=$1 # program is fist becasue for debugging there is not a command provided
```
davesims: "first because" :bowtie:

kyleparrott:
Clearly I need to follow: https://www.linux.com/learn/using-spell-checking-vim 😄

targets/target-stm32f407-disco-gcc/openocd/flash.sh
```diff
@@ -1,27 +1,41 @@
...
+cmd=$2
+if [[ ! -z $cmd ]]; then
+    openocd_arg="$cmd $program"
+fi
 unamestr=`uname`
 
 if [[ "$unamestr" =~ "Linux" ]]; then
-	device=`lsusb -d '0483:'`
+    device=`lsusb -d '0483:'`
 elif [[ "$unamestr" =~ "Darwin" ]]; then
```
plauche: nit: I think we can safely remove this OSX section, as this script is now targeted at running specifically in our vm.

kyleparrott: I still use this natively on my mac, even if just for testing I would like to request that we leave it in there.

kyleparrott:
@davesims: friendly reminder to take another look at this PR.

kyleparrott:
@davesims: Another Friendly bump to take a look at this.

davesims:
whoops, sorry for the delay on this @kyleparrott.

PR #9, opened 2016-11-02
Changing base box to Xenial64 - As it supports the latest version of python out of the box.

kyleparrott:


kyleparrott:
Since the script runs as root, the KubOS source is cloned to `/root/.kubos`. The last command now moves that to `/home/ubuntu/.kubos` where it needs to be as ubuntu is the default user in the box.

PR #12, opened 2016-11-04
[WIP] Telemetry architecture

plauche:
:construction: This PR represents a simple architecture for future telemetry work. :construction: 

At its core telemetry represents a message bus. These messages may represent a variety of data sources (sensors, uptime, system errors) and could be sent to a variety of destinations (beacon, log files, health check, ground station). The goal of this architecture is to decouple the telemetry bus from potential data sources and destinations. 

The telemetry module itself is primarily concerned with:
1. Defining a generic data structure for passing data - **Not currently implemented**
2. Exposing a method for telemetry sources to submit or pass in data
3. Exposing a method for telemetry destinations to retrieve or filter relevant data

The telemetry module should **not** be concerned with:
1. Who is submitting data
2. Who is retrieving data
3. How is that data used, stored or recorded

The current module provides a simple function call for submitting data `telemetry_submit`, a CSP interface for publishing data to telemetry destinations and (will) use a little bit of code generation magic to tie them together. CSP has no good built-in pub-sub functionality, so currently each telemetry destination must create a listening CSP socket and the telemetry module sends all received data over each destination socket. This is currently hardcoded but could be easily generated via config.json.

Telemetry data sources are defined in the config.json file and specify where the data should end up. Code generation turns the destination specifications into flags. Currently all destinations receive all data and must check the destination flag themselves. 

Here is an example of a data source configuration:
```json
{
  "telemetry":{
    "sources":{
      "gps":{
        "beacon":true
      }
    }
  }
}
```
This would indicate a source named gps which should be included in the beacon. This would be translated into the following structure:
```c
static telem_source gps_source = { .source_id = 0, .dest_flag = TELEMETRY_BEACON_FLAG };

```

This data source configuration might also include what type (int/float/char/etc) of data is passed around.

Data destinations such as storage, beacon and health keeping will be responsible for what happens to the data. The telemetry module will not maintain any copies of the data internally. The storage destination will be responsible for recording (all) of the system telemetry. The beacon destination will be responsible for keeping a fresh set its data with AND with sending that out via comms. 

The hope is that this organization of responsibilities will allow us to easily divide up the telemetry ecosystem (sources + base + destinations) into independent tasks :).

/cc @kubostech/core  @kubostech/devs


telemetry/example/scripts/prebuild.py
```diff
@@ -0,0 +1,53 @@
...
+            if 'beacon' in json_data['telemetry']['sources'][source]:
+                dest_flag = "TELEMETRY_BEACON_FLAG"
+                beacon_counter += 1
+            if 'health' in json_data['telemetry']['sources'][source]:
+                if dest_flag == 0:
+                    dest_flag = "TELEMETRY_HEALTH_FLAG"
+                else:
+                    dest_flag += " | TELEMETRY_HEALTH_FLAG"
+                health_counter += 1
+            source_data += source_format % (source, source_counter, dest_flag)
```
marshall: I wonder if we should be using a pre-determined hash, or some other deterministic method for generating `source_id` so JSON changes don't require recompilation in other subsystem projects. 

Also, since the order of keys from a dictionary isn't guaranteed to be the same as the order in JSON file, it might make sense to do something based on the string itself rather than a loop index..

telemetry/example/config.json
```diff
@@ -0,0 +1,23 @@
+{
+    "telemetry":{
+      "beacon":{
+         "interval":2000
+      },
+      "gather_interval":500,
+      "sources":{
+        "gps":{
+          "beacon":true,
+          "health": 100
```
marshall: curious, what do the integer values for `health` represent?

.gitignore
```diff
@@ -59,4 +59,5 @@ upload.tar.gz
 
 # vim files
 .*.sw*
+.vscode
```
marshall: traitor! 🙈

telemetry/telemetry-beacon/source/beacon.c
```diff
@@ -0,0 +1,66 @@
...
+    while(running) {
+        if( (conn = csp_accept(socket, 10000)) == NULL ) {
+            continue;
+        }
+
+        while( (packet = csp_read(conn, 100)) != NULL ) {
+            switch( csp_conn_dport(conn) ) {
+                case TELEMETRY_BEACON_PORT:
+                    data = *((telem_data*)packet->data);
+                    csp_buffer_free(packet);
```
marshall: I know this is PoC code, but you're actually freeing the memory before passing the underlying data to another function here :)

telemetry/telemetry-health/source/health.c
```diff
@@ -0,0 +1,66 @@
...
+#include <csp/csp.h>
+
+static uint8_t data_count = 0;
+
+void update(telem_data data)
+{
+    if (data.source.dest_flag & TELEMETRY_HEALTH_FLAG)
+    {
+        /* A dirty hack to fill up the whole array with data */
+        health_data[data_count++ % TELEMETRY_NUM_HEALTH] = data;
```
marshall: this should definitely be protected by a lock in the real world ;)

telemetry/example/source/main.c
```diff
@@ -0,0 +1,91 @@
...
+    static int pos_data = 2;
+    while(1)
+    {
+        telem_data data;
+        data.data = pos_data--;
+        data.source = pos_y_source;
+        data.timestamp = csp_get_ms();
+
+        telemetry_submit(data);
+        csp_sleep_ms(300);
```
CatFreed: Where does this sleep value come from?  Is it like the gather interval?

telemetry/telemetry-beacon/source/beacon.c
```diff
@@ -0,0 +1,66 @@
...
+#include <csp/csp.h>
+
+static uint8_t data_count = 0;
+
+void telemetry_update_beacon(telem_data data)
+{
+    if (data.source.dest_flag & TELEMETRY_BEACON_FLAG)
+    {
+        /* A dirty hack to fill up the whole array with data */
+        telem_beacon[data_count++ % TELEMETRY_NUM_BEACON] = data;
```
CatFreed: Why are you cycling through beacons?  Why not save data to a specific beacon each time?

telemetry/telemetry-beacon/telemetry-beacon/beacon.h
```diff
@@ -0,0 +1,24 @@
...
+
+#include <stdint.h>
+#include <csp/arch/csp_thread.h>
+
+#include <telemetry/sources.h>
+
+static telem_data telem_beacon[TELEMETRY_NUM_BEACON];
+
+CSP_DEFINE_TASK(beacon_thread);
+CSP_DEFINE_TASK(beacon_rx_thread);
```
CatFreed: A comment blob would be nice to explain the difference between these.  Is beacon_rx for peripheral->OBC and then beacon for OBC->ground?

telemetry/telemetry-beacon/telemetry-beacon/beacon.h
```diff
@@ -0,0 +1,24 @@
...
+
+#ifdef YOTTA_CFG_TELEMETRY
+#define BEACON_THREAD   csp_thread_handle_t beacon_handle; \
+                        csp_thread_handle_t beacon_rx_handle; \
+                        csp_thread_create(beacon_thread, "BEACON", 50, NULL, 0, &beacon_handle); \
+                        csp_thread_create(beacon_rx_thread, "BEACON_RX", 50, NULL, 0, &beacon_rx_handle);
+#else
+#define BEACON_THREAD
```
CatFreed: nit: BEACON_THREADS would be better, since you're creating multiple threads

telemetry/telemetry/source/telemetry.c
```diff
@@ -0,0 +1,62 @@
...
+
+void telemetry_submit(telem_data data)
+{
+    telemetry_log_data(data);
+    telemetry_publish(data);
+}
+
+void telemetry_log_data(telem_data data)
+{
+    printf("TELEM:%d:%d:%d\r\n", data.source.source_id, data.timestamp, data.data);
```
CatFreed: Would we be saving to a log file as well some time in the future?

telemetry/example/config.json
plauche: These values are not currently used, but the thinking was the user could health limits for data sources. The "health" service could verify that the telemetry data is within the specified limits.

telemetry/telemetry-beacon/source/beacon.c
plauche: The beacon is actually made up of all the data in the array. We are just cycling through the different data points in the beacon.

plauche:
@CatFreed The beacon and health aren't too different right now because I didn't want to dive too deep into them. The idea right now was more to show two different destinations for the telemetry data.

In the future the "health" destination could be a limit checker of sorts, while the "beacon" is the heartbeat of specified data to the ground.

plauche:
This PR will be kept around for information purposes but will not be built on for actual usage. A new PR (https://github.com/kubostech/kubos/pull/14) has been opened with the core telemetry layer for creating a working starting point.

kyleparrott:
@plauche: since #14 has been merged can we close this issue or do we need to keep it open for a while longer?

PR #16, opened 2016-11-09
Telemetry Data Type

cnalley:
This PR adds a union data type to the telemetry packet struct and an enum for telemetry data types to allow destinations to decode which union member is being used. This is intended to start the discussion on the pros and cons of using a union to store telemetry data.

/cc @kubostech/devs

### DIFF
--- a/telemetry/telemetry/telemetry.h
+++ b/telemetry/telemetry/telemetry.h
@@ -48,3 +48,4 @@ typedef struct
 void telemetry_submit(telemetry_packet packet);
 
 #endif
+
--- a/telemetry/telemetry/telemetry.h
+++ b/telemetry/telemetry/telemetry.h
@@ -19,7 +19,7 @@ typedef struct
 typedef enum {
     TELEMETRY_TYPE_INT = 0,
     TELEMETRY_TYPE_FLOAT
-} telemetry_type;
+} telemetry_data_type;
 
 /**
  * Telemetry union for data.
@@ -28,7 +28,7 @@ typedef union
 {
     int i;
     float f;
-} telem_union;
+} telemetry_union;
 
 /**
  * Basic telemetry packet structure - encapsulating routing information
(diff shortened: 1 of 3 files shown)


telemetry/telemetry/telemetry.h
```diff
@@ -4,23 +4,41 @@
...
 
 /**
- * Telemetry packet routing information structure
+ * Telemetry packet routing information structure.
  */
 typedef struct
 {
     uint8_t source_id;
     uint8_t dest_flag;
+    uint8_t data_type;
```
plauche: I'm torn between the idea of storing data type here in the `telemetry_source` struct and storing the type adjacent to the data in the packet struct.

telemetry/telemetry/telemetry.h
```diff
@@ -4,23 +4,41 @@
...
 } telemetry_source;
 
 /**
+ * Telemetry data types.
+ */
+typedef enum {
+    TELEMETRY_TYPE_INT = 0,
+    TELEMETRY_TYPE_FLOAT
+} telemetry_type;
```
plauche: Maybe this should be called `telemetry_data_type`.

telemetry/telemetry/telemetry.h
```diff
@@ -4,23 +4,41 @@
...
+
+/**
+ * Telemetry union for data.
+ */
+typedef union
+{
+    int i;
+    float f;
+} telem_union;
```
plauche: nit: Let`s fully spell out telemetry

telemetry/telemetry/telemetry.h
cnalley: The thought behind storing the data type in the source struct was to keep that struct seperate for a code generation tool to create sources based on a type declared in the config.json. Although, putting it in the packet struct would make it more convenient to access during the many "type checks" that will be performed by other code.

PR #18, opened 2016-11-17
Basic aggregator for telemetry

plauche:
A basic data aggregation layer for telemetry
* Currently no automatic data polling
* Provides a stub function for users to implement with their sensor polling/data submission
* User function is run in loop with user configurable delay
* Provides helper function for telemetry submission 

CC: @kyleparrott @cnalley @kubostech/devs

### DIFF
--- a/telemetry-aggregator/source/aggregator.c
+++ b/telemetry-aggregator/source/aggregator.c
@@ -1,0 +1,25 @@
+#ifdef YOTTA_CFG_TELEMETRY_AGGREGATOR
+
+#include "telemetry-aggregator/aggregator.h"
+#include <csp/csp.h>
+
+CSP_DEFINE_TASK(aggregator)
+{
+    while(1)
+    {
+        user_aggregator();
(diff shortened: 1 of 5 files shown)


telemetry-aggregator/source/aggregator.c
```diff
@@ -0,0 +1,25 @@
...
+
+#include "telemetry-aggregator/aggregator.h"
+#include <csp/csp.h>
+
+CSP_DEFINE_TASK(aggregator)
+{
+    while(1)
+    {
+        user_aggregator();
+        csp_sleep_ms(YOTTA_CFG_TELEMETRY_AGGREGATOR_INTERVAL);
```
kyleparrott: Is there a place we are documenting the options that can be configured in the project's config.json file yet?

plauche: I haven't been documenting them yet, but that would be good to do going forward.

PR #19, opened 2016-11-28
Add build option to libcsp to allow RDP to be set in a projects config.json

kyleparrott:


plauche:
Looks good :shipit: :rocket:

PR #20, opened 2016-11-28
KUBOS-206 STM32F4 HAL changes for receiving CSP packets over UART

kyleparrott:
Plug the usart_callback back into the UART irq for the STM HAL to allow CSP packets to be received over UART.


kyleparrott:
This is not the most elegant solution, and is not reusable across the msp430 HAL. [KUBOS-259](https://kubostech.atlassian.net/browse/KUBOS-259) has been created to address this at a later time.

kyleparrott:
This PR was not needed. The CSP uart handler that called the usart callback is already implemented in:  https://github.com/kubostech/kubos/blob/master/libcsp/src/drivers/usart/usart_kubos.c#L28-L41

PR #26, opened 2016-12-06
[WIP] Created jenkinsnode branch; added working code

jessehamner:
Added working python and shell scripts. At present, the flash.sh is incompatible with the existing flash.sh, BUT it shouldn't be hard to change -- target.json should do it.


test/integration/jenkinsnode/cistack.py
```diff
@@ -0,0 +1,408 @@
...
+
+    thepin=pinvals[key][0]
+
+    if (thepin is False):
+        return 1
+    if (thepin == 'board'):
+        return 1
+
+    try:
+        func=GPIO.gpio_function(thepin)
```
plauche: nit: We should be consistent in our spacing around operators, generally we have been adding one space on either side.

test/integration/jenkinsnode/cistack.py
```diff
@@ -0,0 +1,408 @@
...
+
+    arguments = parser.parse_args()
+    args = vars(arguments)
+
+    return args
+
+
+def flashBinary(binfile,binpath, flashpath="/usr/local/lib/python2.7/dist-packages/kubos/flash/openocd",  **pinvals):
+    '''This function determines which board is in use, and flashes the binary using an appropriate binary executable and/or shell script(s).'''
+    if (pinvals['board'] == "pyboard-gcc"):
```
plauche: There are a few different places where we branch based on the board. These are going to start adding up as we add support for more targets. I would suggest abstracting these common functions away into a `Target` class to clean this up.

test/integration/jenkinsnode/cistmdiscotest.py
```diff
@@ -0,0 +1,84 @@
+#!/usr/bin/python
```
plauche: This file name could use some work. Maybe we could have a tests folder and inside a `stm32f4disco.py` and `pyboard.py`.

test/integration/jenkinsnode/flash.sh
```diff
@@ -0,0 +1,59 @@
+#!/bin/bash
```
plauche: Will this eventually be migrated into the yotta target and referenced from there?

test/integration/jenkinsnode/cistmdiscotest.py
jessehamner: Totally fair. I'll change them.

test/integration/jenkinsnode/flash.sh
jessehamner: That's one option -- parsing the ```flash.sh``` script arguments was wonky or broken on ARM debian, so these changes would probably add a bit of robustness. Since you told me that the JSON file might be the only other required change...yes, I would like to at least consider merging this version of the script once I confirm the JSON change works as expected.

test/integration/jenkinsnode/LICENSE.txt
```diff
@@ -0,0 +1,203 @@
+
```
plauche: @marshall @davesims is it necessary to place a LICENSE.txt file in each new module now that all modules are in the same repo?

test/integration/jenkinsnode/cistack.py
plauche: To be clear I meant one space on each side. You were already doing this with your comparison operators, just not with the assignment ones.

test/integration/jenkinsnode/LICENSE.txt
jessehamner: I did this without knowing that answer -- but it seemed prudent compared to leaving it out.

test/integration/jenkinsnode/cistack.py
plauche: I'm fine with pushing off this specific refactor into a new task (https://kubostech.atlassian.net/browse/KUBOS-270) to be worked after this branch is merged.

test/integration/jenkinsnode/flash.sh
plauche: Ok it will be fine here for now. We should have a talk about how many more of these scripts you'll need. If you will always be flashing in the same way that we flash with yotta/kubos-cli, then we should investigate how to consolidate the scripts.

jessehamner: I will actually leave the flash scripts untouched, and implement the work they do in a python function instead. That's some minor duplication of effort, but it's not bad and it's cleaner code with fewer moving parts and handoffs.

jessehamner:
I'm requesting to delete several files, as they have been superseded by (a) functions in cistack.py and (b) a more generic python test script.

jessehamner:
This branch is separate from the "provisioning" branch, that uses a big shell script and some ansible playbooks. I'll create a separate branch and PR for those files.

test/integration/jenkinsnode/cistack.py
```diff
@@ -0,0 +1,542 @@
...
+import datetime
+
+def allDone():
+    '''Free all the pins and exit the script.'''
+    GPIO.cleanup()
+    sys.exit("Pins cleared. Exiting script.")
+    return 1
+
+
+def pinSetup(key, **pinvals):
```
plauche: This function could be simplified by passing in just `thepin`, instead of the key and the whole pin array.

test/integration/jenkinsnode/cistack.py
```diff
@@ -0,0 +1,542 @@
...
+    if (pinvals[key][0] is False):
+        return 1
+    try:
+        GPIO.output(pinvals[key][0], pinvals[key][2])
+    except:
+        return 0
+    return 1
+
+
+# Setting BCM mode is "Broadcom", running from GPIO2 to GPIO27
```
jessehamner: Done

test/integration/jenkinsnode/cistack.py
```diff
@@ -0,0 +1,542 @@
...
+def pinOn(key, **pinvals):
+    '''Generic "assert the GPIO pin" function.'''
+    if (pinvals[key][0] is False):
+        return 1
+    try:
+        print("Asserting pin %s" % str(pinvals[key][0])   )
+        GPIO.output(pinvals[key][0], pinvals[key][1])
+        
+    except:
+        print("ERROR: unable to assert pin %s" % str(pinvals[key][0])  )
```
jessehamner: Done. Old habits die hard.

test/integration/jenkinsnode/cistack.py
```diff
@@ -0,0 +1,542 @@
...
+
+def pinLayout(target):
+    ''' Configure the board-specific pin addresses and directions '''
+    if (target == "pyboard-gcc"):
+        pinvals = {
+        'board':target,
+        'rst':[17, True, False, GPIO.OUT], 
+        'prg':[18, True, False, GPIO.OUT],
+        'pwr':[27, True, False, GPIO.OUT],
+	    'opt':[22, True, False, GPIO.OUT]
```
jessehamner: Stray tabs! fixed.

test/integration/jenkinsnode/cistack.py
```diff
@@ -0,0 +1,542 @@
...
+import serial
+import spidev
+import argparse
+import datetime
+
+def allDone():
+    '''Free all the pins and exit the script.'''
+    GPIO.cleanup()
+    sys.exit("Pins cleared. Exiting script.")
+    return 1
```
jessehamner: Done. Will test.

test/integration/jenkinsnode/cistack.py
```diff
@@ -0,0 +1,542 @@
...
+
+    return 1
+
+def readOpts():
+    args= { 'board': 'stm32f407-disco-gcc',
+            'binary': 'kubos-rt-example',
+            'path': '/var/lib/ansible/'
+            }
+
+    try:
```
jessehamner: Done.

test/integration/jenkinsnode/LICENSE.txt
davesims: Starting a new comment/review, so GH will recognize this is a request for changes (sigh). Let's ditch this license file and rely on the one at the root of the repo.

test/integration/jenkinsnode/cistack.py
```diff
@@ -0,0 +1,731 @@
...
+    if (pinvals[key][0] is False):
+        return True
+    try:
+        GPIO.output(pinvals[key][0], pinvals[key][2])
+    except:
+        return False
+    return True
+
+
+def setupBoard(**pinvals):
```
davesims: Has this `setupBoard` been deprecated by https://github.com/kubostech/kubos/pull/26/files#diff-93554954731cc54bd272a61e74cd4b01R171 ?

test/integration/jenkinsnode/LICENSE.txt
jessehamner: Probably. I'll remove it.

test/integration/jenkinsnode/cistack.py
```diff
@@ -0,0 +1,731 @@
...
+            pass
+        else:
+            pinSetup(pin, **pinvals)
+
+    print("Setup successful.")
+    sleep(0.1)
+    return True
+
+
+def determineFileType(binfile):
```
davesims: So many operations on binfile (`determinaFileType`, `pyBoardSanityCheck`, etc.) suggest to me it needs its own class. What do you think @plauche @jessehamner ?

test/integration/jenkinsnode/cistmdiscotest.py
```diff
@@ -0,0 +1,84 @@
...
+import os
+from time import sleep
+import RPi.GPIO as GPIO
+import cistack as ci
+import subprocess
+
+pinvals = ci.pinLayout('stm32f407-disco-gcc')
+p=pinvals
+
+flashloc="/usr/local/lib/python2.7/dist-packages/kubos/flash/"
```
jessehamner: I'm fine getting those paths from ENV variables; We'll have to declare them in the provisioning process though -- kubos-sdk won't fully install on RPi. 

*Also*: these scripts have been replaced wholesale by a single `boardflash.py` file.

test/integration/jenkinsnode/cistack.py
jessehamner: I can do that.

test/integration/jenkinsnode/cistack.py
jessehamner: As requested, made a `Pin` class

test/integration/jenkinsnode/cistack.py
```diff
@@ -0,0 +1,542 @@
...
+    return 1
+
+
+def powerDown(**pinvals):
+    '''Turn off the power MOSFET for the target board.'''
+    pinOff('pwr', **pinvals)
+    return 1
+
+
+def pinOn(key, **pinvals):
```
jessehamner: New `Pin` class should solve this?

test/integration/jenkinsnode/cistmdiscotest.py
jessehamner: I'm getting almost all of these from ENV variables now. One I still have to construct from a known ENV variable (KUBOS_LIB_PATH) and jump around to the `../../flash/openocd` directory.

test/integration/jenkinsnode/cistmdiscotest.py
```diff
@@ -0,0 +1,84 @@
...
+shutdownwhendone=0
+ignoreGPIOwarnings=False
+
+if (ignoreGPIOwarnings):
+    GPIO.setwarnings(False)
+
+
+ci.setupBoard(**p)
+print ("Board setup complete.")
+sleep(1)
```
jessehamner: Yep -- I don't want to start manipulating pins until everything is nice and settled. I figure I only need to wait 100ms though at most.

test/integration/jenkinsnode/cistmdiscotest.py
```diff
@@ -0,0 +1,84 @@
...
+
+# should parse options here
+# but instead:
+binfile = "disco-ukub-sensor-node"
+binpath = "/home/kubos"
```
jessehamner: On the one hand -- this is only intended to run on a RasPi; on the other, I've been implementing ENV variables.

test/integration/jenkinsnode/tests/boardflash.py
```diff
@@ -0,0 +1,96 @@
...
+from time import sleep
+import RPi.GPIO as GPIO
+import cistack as ci
+import subprocess
+
+kwargs = ci.readOpts()
+
+print(str(kwargs))
+
+# 'inputbinary': 'kubos-rt-example', 'binfilepath': '/home/kubos', 'verbose': True, 'board': 'stm32f407-disco-gcc'
```
jessehamner: Yep. The whole script has been deleted now and incorporated into `ci.py.`

test/integration/jenkinsnode/cistack.py
```diff
@@ -0,0 +1,731 @@
...
+        '''Turn off the power MOSFET for the target board.'''
+        pinOff('pwr', **self.pins)
+        return True
+
+    def setupboard(self):
+        '''Run this function immediately after determining which pins
+are assigned to your target board. This function sets the Raspberry
+Pi GPIO map to "Broadcom" and then sets up pin direction and function.'''
+
+#    print("Checking for root access...")
```
jessehamner: Fixed.

test/integration/jenkinsnode/cistack.py
```diff
@@ -0,0 +1,731 @@
...
+        for pin in self.pins.keys():
+            pinSetup(pin, **self.pins)
+
+        print("Setup successful.")
+        sleep(0.1)
+        return True
+
+
+
+class Pyboard(Target):
```
jessehamner: Lots more in this most recent PR.

test/integration/jenkinsnode/cistack.py
```diff
@@ -0,0 +1,731 @@
...
+        command = str("%s -f %s/%s -s %s -c \"%s %s/%s\"") % (openocdloc, 
+            searchpath, cfg, searchpath, cmd, binpath, binfile)
+        print (str(command))
+        try:
+            subprocess.check_output(command, shell = True)
+            return True
+        except:
+            return False
+
+def pinSetup(key, **pinvals):
```
jessehamner: Yep.

test/integration/jenkinsnode/cistack.py
jessehamner: *DONE.*

jessehamner:
Lots of OO work this last PR. It still has some noise but it is rapidly improving.

test/integration/jenkinsnode/cistack/utils.py
```diff
@@ -0,0 +1,366 @@
...
+    try:    
+        retval = subprocess.check_output(cmd, shell = True)
+        retval = re.sub('\n$', '', retval)
+        return retval
+    except:
+        sys.exit(str("Unable to determine the path to %s; halting." % 
+            command))
+        return False
+
+def checkRoot(yesno):
```
plauche: This boolean flag parameter would be better used to determine whether to call this function. It does not really serve a purpose inside of the function, except to decide if the function's code should be executed or not.

test/integration/jenkinsnode/ci.py
```diff
@@ -0,0 +1,126 @@
...
+import magic 
+import cistack as ci
+
+def main():
+
+    kwargs = ci.readOpts()
+    print(str(kwargs))
+
+    global binfile, binpath, board, isrootrequired
+    binfile = kwargs['inputbinary']
```
plauche: Again, not sure why we are turning the arguments namespace into a dictionary, only to store the dictionary elements in their own variables. The namespace that `parse_args` returns could be used directly here instead of these variables.

test/integration/jenkinsnode/ci.py
```diff
@@ -0,0 +1,126 @@
...
+    requiredutils = ci.requiredUtils()
+
+    if ci.sanityChecks(*requiredutils):
+        print("Located required system utilities. Continuing.")
+    else:
+        sys.exit("Unable to locate required utilites. Exiting.")
+
+    pathkeys = ci.requiredPaths()
+    pathdict = dict.fromkeys(pathkeys, "")
+    if ci.pathChecks(**pathdict):
```
plauche: Why does `ci.pathChecks()` need to have a dictionary instead of the array that `ci.requiredPaths()` returns?

test/integration/jenkinsnode/ci.py
```diff
@@ -0,0 +1,126 @@
...
+
+        target.reset()
+        sleep(1)
+
+        target.progmode()
+        sleep(1)
+
+        if (target.flash(b) is True):
+            print("Program flash completed. Reports success.")
+        else:
```
plauche: Do we really want to exit immediately if the flash fails? It seems like it would still be appropriate to perform cleanup (reset board, shut down, free pins) in this case.

test/integration/jenkinsnode/cistack/binfile.py
```diff
@@ -0,0 +1,182 @@
...
+# try very hard to find the binary
+        if not self.getpath():        
+            sys.exit("%s unable to find binary file to upload in \
+            specified path or current working directory %s. \
+            Exiting now." % (errstr, str(array[0])))
+
+# try moderately hard to figure out binary file type / arch
+        array = self.getfiletype()
+        if not (array[0] or array[1]):
+            raise IOError
```
plauche: It is not clear why we are raising an `IOError` here instead of returning false or exiting.

test/integration/jenkinsnode/cistack/binfile.py
```diff
@@ -0,0 +1,182 @@
...
+    def path(self):
+        return self.path
+
+    def board(self):
+        return self.board
+
+    def name(self):
+        return self.name
+
+    def file(self):
```
plauche: Do we need two functions which both return `self.name`?

test/integration/jenkinsnode/cistack/pin.py
```diff
@@ -0,0 +1,109 @@
...
+    def setup(self):
+        """Set up one GPIO pin per the pin dict values. """
+        
+        GPIO.setmode(GPIO.BCM)
+
+        if self.number is False:
+            return False
+        if self.number is None:
+            return False
+        if not self.direction == GPIO.OUT:
```
plauche: This would be more cleanly stated `if self.direction != GPIO.OUT`

test/integration/jenkinsnode/cistack/utils.py
```diff
@@ -0,0 +1,366 @@
...
+
+
+def getTarget(target):
+    from nasatbus import NAsatbus
+    from pyboard import Pyboard
+    from stm32f407discovery import STM32F407Discovery
+    from msp430f5529 import MSP430
+
+    """ Configure the board-specific pin addresses and directions """
+    print("Checking on the availability of the %s Target class" % target)
```
plauche: We may want to consider transitioning all of this `print` output onto the logging module. Using logging will give us a bit more fine grained control over how/when this output is shown.

test/integration/jenkinsnode/cistack/msp430f5529.py
```diff
@@ -0,0 +1,77 @@
...
+            'pwr' : Pin(name = 'pwr', number = 27),
+            'opt' : Pin(name = 'opt', number = 22)  # optional
+        }
+    
+    def flash(self, binobj):
+        from utils import findBin
+        """Flash a binary file through USB connection on an MSP430 Launchpad"""
+
+        if not self.sanitycheck(binobj):
+            sys.exit("Binary file didn't pass a sanity check. Exiting.")
```
plauche: I know we are found of using `sys.exit` to fail fast but it may be worth considering removing these exits and returning back errors. If we bail with an exit there is no chance to cleanup or reset things to a nice state (reset board, shut down board, free pins).

test/integration/jenkinsnode/cistack/utils.py
```diff
@@ -0,0 +1,366 @@
...
+requires root access to run, meaning you must implement \
+a udev rule to enable user-land access to the device, or else just \
+remember to sudo when running this script. \n\
+\n\
+Hint: ATTRS{idVendor}==\"0483\", ATTRS{idProduct}==\"df11\" \n\
+\n\
+Therefore, you can set this to True, but we don't advise it.", \
+        metavar = "ROOT", required = False)
+
+    parser.add_argument("-f", "--file", action = 'store', \
```
jessehamner: Yes, but there is some overlap. If `-p`, it specifically declares a path for the binary executable. If `-f`, the program checks for a path and splits the binfile off of it.

test/integration/jenkinsnode/cistack/utils.py
```diff
@@ -0,0 +1,366 @@
...
+        help = "the target board (and architecture) supported by the \
+Kubos SDK", metavar = "TARGET", required = True)
+   
+    parser.add_argument("-i", "--ignore-warnings", action = 'store', \
+        dest = 'ignoreGPIOwarnings', default = False, \
+        help = "Ignore any warnings from the RPi.GPIO module. Not \
+recommended.",
+        metavar = "IGNOREWARNINGS", required = False)
+
+    parser.add_argument("--free-pins", action = 'store', \
```
jessehamner: Yes. Freeing the pins when done is _good_ behavior, though in practice it does not matter.

test/integration/jenkinsnode/cistack/utils.py
```diff
@@ -0,0 +1,366 @@
...
+        dest = 'freepins', default = False, \
+        help = "Use RPi.GPIO module to Free GPIO pins when done. Usually \
+unnecessary.",
+        metavar = "FREEPINSWHENDONE", required = False)
+
+    parser.add_argument("--shutdown", action = 'store', \
```
jessehamner: As above, yes, though it doesn't matter much. Typically we would leave the board running so it could run unit tests and provide hardware checks, e.g.

test/integration/jenkinsnode/cistack/utils.py
```diff
@@ -0,0 +1,366 @@
...
+        dest = 'command', default = "flash", type = str, \
+        nargs = '?', const = 'flash', \
+        help = "Assign a functionality to the current session. \n\
+        - 'flash' is default; \n\
+        - 'lib' imports the file as a library; \n\
+        - 'checks' runs tests on each of the flags and options.\n",\
+        metavar = "COMMAND", required = False)
+
+    arguments = parser.parse_args()
+    args = vars(arguments)
```
jessehamner: Nope. Guess I'll read more about argparse.

test/integration/jenkinsnode/ci.py
jessehamner: Yeah, it's wasteful to do it this way. It allowed me to have more concise syntax, but that's about all.

test/integration/jenkinsnode/ci.py
jessehamner: Originally I figured it might be good to populate the dict values with `os.environ[]` values for each environment variable (key), but in retrospect that seems like a weak choice. I'll fix it.

test/integration/jenkinsnode/ci.py
jessehamner: Good point. Will fix.

test/integration/jenkinsnode/cistack/binfile.py
jessehamner: Is it preferable to just throw an `except`? I thought specificity was better, but it might be clouding the issue.

test/integration/jenkinsnode/cistack/utils.py
plauche: :) I ask about this and the board shutdown mainly to call out that any usage of `sys.exit` negates these options.

test/integration/jenkinsnode/cistack/binfile.py
plauche: Well I'm not sure this is the best place to throw an exception. None of the calls to this function catch any exceptions, so this would immediately take down your program. It probably would be better to return False, though none of the calls to this function check the return value. You could just set `self.arch` and `self.filetype` from inside of `self.getfiletype()`, since that is a class method. It could then return True/False depending on whether a valid file was found or not.

test/integration/jenkinsnode/cistack/binfile.py
```diff
@@ -0,0 +1,182 @@
...
+    def abspath(self):
+        return os.path.join(self.path,self.name)
+
+    def arch(self):
+        return self.arch
+
+    def filetype(self):
+        return self.filetype
+
+# Big effort to find path to binary:
```
davesims: This is a general note but not a request for change -- it's fine like it is, but I think the conditions you're commenting on here are pretty self-documenting. A rule of thumb for me is, I comment only if the code or the reasons behind it aren't transparent. Over time comments & code tend to get out of sync.

test/integration/jenkinsnode/cistack/utils.py
jessehamner: Making progress here. Lots of changes to drive errors to call the `cleanUp()` function instead of just calling `sys.exit()` immediately.

test/integration/jenkinsnode/cistack/binfile.py
jessehamner: Fixed

jessehamner:
Migrating the `print` statements to `logging` library calls. Enabling more graceful shutdown of the script in the face of errors. Still WIP but progressing.

test/integration/jenkinsnode/cistack/pin.py
jessehamner: Fixed.

test/integration/jenkinsnode/cistack/binfile.py
```diff
@@ -0,0 +1,183 @@
...
+    """
+    A binary file object that contains the name, file type, path, 
+    target architecture, and target board type. Internal methods 
+    include file and path validation, attempts to determine target
+    architecture and file type, and a series of variable return
+    methods.
+    """
+
+    def __init__(self, name = '', path = '', board = ''):
+        self.board = board
```
plauche: What is the purpose of the board variable in this class? It looks like it is just validated & printed but not used for any logic/work.

test/integration/jenkinsnode/ci.py
```diff
@@ -0,0 +1,133 @@
...
+import argparse
+import datetime
+import magic 
+import cistack as ci
+
+def main():
+
+    args = ci.readOpts()
+
+#    log = logging.getLogger('logfoo')
```
plauche: nit: Let's cleanup this commented code

test/integration/jenkinsnode/ci.py
```diff
@@ -0,0 +1,133 @@
...
+        logging.info("\nBoard reset.")
+
+        sleep(1)
+
+# If you want to shut the board down, this command cleans up and 
+# de-energizes the power MOSFET, and frees the GPIO pins, if flagged
+# to do so.
+        ci.cleanUp(target, args)
+
+    elif args.command == "lib":
```
plauche: Nit: It would be better for now to leave out un-implemented commands, rather than having unused code floating around.

test/integration/jenkinsnode/ci.py
```diff
@@ -0,0 +1,133 @@
...
+        pins = target.getpins()
+        logging.debug(str(pins))
+
+        target.powerup()
+        sleep(0.5)
+
+        target.reset()
+        sleep(0.5)
+
+        target.progmode()
```
plauche: Should we check the return value of setupboard/powerup/reset/progmode before attempting to flash?

test/integration/jenkinsnode/cistack/stm32f407discovery.py
```diff
@@ -0,0 +1,95 @@
...
+            searchpath, cfg, searchpath, cmd, fileloc)
+        print (str(command))
+        try:
+            subprocess.check_output(command, shell = True)
+            return True
+        except:
+            return False
+
+
+    def sanitycheck(self, binobj):
```
jessehamner: Fixed

test/integration/jenkinsnode/cistack/utils.py
jessehamner: WIP but comments welcome.

test/integration/jenkinsnode/cistack/utils.py
```diff
@@ -0,0 +1,366 @@
...
+    """Ensure that the board identifier is supported."""
+    for i in boards:
+        try:
+            r = parseBoardIdentifier(i['dev'])
+            if(r[1] is True): # board is supported
+                return r
+        except:
+            sys.exit("Unable to determine board type. Exiting.")
+
+    return False
```
jessehamner: Done.

test/integration/jenkinsnode/cistack/utils.py
jessehamner: Fixed.

test/integration/jenkinsnode/cistack/binfile.py
jessehamner: Is using it for validation not enough to justify its inclusion?

test/integration/jenkinsnode/ci.py
jessehamner: No problem.

test/integration/jenkinsnode/cistack/utils.py
```diff
@@ -0,0 +1,445 @@
...
+    return Basicbinfile(name = name, path = path, board = board)
+
+
+def getTarget(target):
+    from nasatbus import NAsatbus
+    from pyboard import Pyboard
+    from stm32f407discovery import STM32F407Discovery
+    from msp430f5529 import MSP430
+
+#     log = logging.getLogger('logfoo')
```
jessehamner: Fixed, I think.

test/integration/jenkinsnode/cistack/utils.py
```diff
@@ -0,0 +1,445 @@
...
+
+
+def findBin(command):
+#    log = logging.getLogger('logfoo')
+    cmd = str("/usr/bin/which %s" % command)
+    logging.debug("Looking for %s in system binary PATHs." % command)
+
+    try:    
+        retval = subprocess.check_output(cmd, shell = True)
+        retval = re.sub('\n$', '', retval)
```
jessehamner: Done (used `rstrip`)

test/integration/jenkinsnode/cistack/utils.py
```diff
@@ -0,0 +1,445 @@
...
+        print("You need to have root privileges to run this script.\n")
+        sys.exit("Please try again, this time using 'sudo'. Exiting.")
+        return False
+    else:
+        return True
+
+
+def readOpts():
+    """Read command line arguments and return Namespace object."""
+    today = datetime.date.today()
```
jessehamner: Correct! fixed.

test/integration/jenkinsnode/cistack/utils.py
```diff
@@ -0,0 +1,445 @@
...
+    if board in supportedboards:
+        return True
+    logging.error(str("%s Board name '%s' does not match list of currently supported \
+boards. Exiting." % (errstr, board)))
+    return False
+
+def allDone():
+    """Free all the pins and exit the script."""
+    GPIO.cleanup()
+    sys.exit("Pins cleared. Exiting script.")
```
jessehamner: Ah, yep. Fixed.

test/integration/jenkinsnode/cistack/utils.py
```diff
@@ -0,0 +1,445 @@
...
+    Check for dfu-util, openocd, mspdebug, portions of the Kubos SDK,
+    and other stuff.
+    """
+#    log = logging.getLogger('logfoo') 
+    
+
+    logging.debug("Command line arguments are: %s " % str(args))
+    if not args:
+        logging.error("Command line arguments returned None. Exiting.")
+        return None
```
jessehamner: Done.

test/integration/jenkinsnode/cistack/utils.py
```diff
@@ -0,0 +1,445 @@
...
+                logging.error(str("No environment variable %s" % i))
+                return False
+        except:
+            logging.error(str("No environment variable %s" % i))
+            return False
+
+    return True
+
+
+def getEnvironmentVariables(requiredpaths):
```
jessehamner: It was quickly superseded by the tidier function `pathChecks()`

test/integration/jenkinsnode/cistack/target.py
```diff
@@ -0,0 +1,117 @@
...
+    def getboard(self):
+        return self.board
+
+    def getpins(self):
+        return self.pins
+
+    def arch(self):
+        return self.arch
+
+    def processor(self):
```
jessehamner: Fixed.

test/integration/jenkinsnode/cistack/target.py
```diff
@@ -0,0 +1,117 @@
...
+        self.pins['pwr'].on()
+        sleep(0.1)
+        return True
+
+    def powerdown(self):
+        """Turn off the power MOSFET for the target board."""
+#        log = logging.getLogger('logfoo')
+        logging.debug("Turning off the board's power")
+        self.pins['pwr'].off()
+        return True
```
jessehamner: Implemented for `pwr` and `rst`.

test/integration/jenkinsnode/cistack/binfile.py
plauche: The main issue is that we are tightly coupling binfile validation with the board, when the binfile shouldn't really be concerned with the board type. A binfile does not have a board and does not change behavior based on the board type. 

We also already validate the board type when attempting to create the target object, which is dependent on the board type.

plauche:
I'd suggest that we get this PR merged in before adding any additional functionality (i.e. support for the nano avionics board).

plauche:
A friendly nudge @jessehamner to merge this in soon, this PR is approaching 2 months in age

jessehamner:
Done. Thanks for the nudge.

plauche:
This branch has been updated with the latest commits from master and should now be conflict free.

PR #27, opened 2016-12-12
[KUBOS-334] SDK target for Linux on iOBC

CatFreed:
I have created a new target to build a user application to run on the iOBC.  In order to work, it requires that the BuildRoot toolchain be present on the build system at /usr/bin/iobc_toolchain.  (This can be done by building Linux with BR2_HOST_DIR=/usr/bin/iobc_toolchain)


plauche:
Can we make this (and buildroot) work with an arm linux toolchain that we bundle with the sdk?

plauche:
We might also want to consider naming this target `target-kubos-linux-isis-gcc` if we are intending to port RT to this platform.

CatFreed:
@plauche 
Ideally, yes, we can bundle the toolchain with the SDK, but I'm not sure what needs to be done to make that happen.  The buildroot toolchain can be exported by setting the BR2_HOST_DIR variable, which tells buildroot where to put the toolchain during buildtime.  Whatever folder is specified can then be copied and used elsewhere.  The toolchain can then be imported and used with the BR2_TOOLCHAIN_EXTERNAL_PATH variable.

There are a couple catches:
1. I had issues copying the folder (/opt/buildroot) out of my VM because of issues with VirtualBox and relative links.  I'm assuming that that can be resolved, but I don't know where the problem lies (good chance for user-error...) or what kind of work-effort would be required.
2. The buildroot toolchain must use an absolute file path.  If I build the toolchain in /opt/buildroot, it better be copied to /opt/buildroot on my end-point machine.

plauche:
I was thinking we would use a toolchain like this one (http://packages.ubuntu.com/search?keywords=gcc-arm-linux-gnueabi) that could be installed in the sdk vm pre-distribution. @kyleparrott can assist with getting that integrated in. Surely buildroot has the ability to point to an existing external toolchain.

CatFreed:
The kubos-linux-isis-gcc target will now use the buildroot toolchain in /usr/bin/iobc_toolchain.  This toolchain has been added to the vagrant image in my kubos-vagrant PR.

PR #29, opened 2017-01-03
[KUBOS-277] Telemetry testing

plauche:
* The communications.c/h files from telemetry have been moved into their own `ipc` module, and they have been rename `pubsub.c/h`, as that is the specific type of ipc functionality they provide.

* Unit testing has been implemented against both the `telemetry` and `ipc` modules. Tests are implemented using the [`cmocka`](https://cmocka.org/) library, using mocks to stub out dependent functions. This was especially useful in testing telemetry/ipc as the real dependencies tend to live in other threads and can be hard to replicate in a testing environment. The api for using cmocka can be found [here](https://api.cmocka.org/).

Here is the basic structure for the test files/folders:

telemetry
|-source
|-telemetry
|-test - **Test folder by yotta convention**
|---telemetry - **A subfolder is created to compile multiple files into a single test**
|-----[telemetry.c](https://github.com/kubostech/kubos/blob/0bae601aacfdfbc6dc94f47cc466d68d4ede2af4/telemetry/test/telemetry/telemetry.c) - **Contains the actual tests**
|-----[ipc.c](https://github.com/kubostech/kubos/blob/0bae601aacfdfbc6dc94f47cc466d68d4ede2af4/telemetry/test/telemetry/ipc.c) - **Contains stub/mock functions used by tests. In this case the file is called "ipc" because that is the dependency which is stubbed**
|-----[stubs.cmake](https://github.com/kubostech/kubos/blob/0bae601aacfdfbc6dc94f47cc466d68d4ede2af4/telemetry/test/telemetry/stubs.cmake) - **Contains special linking instructions to insert stubs/mocks**

Stub/mock functions are currently stored in a source file in the same folder as the test file. Yotta requires both of these files to live in a folder underneath test if they are to be compiled into the same test binary. 

The stub/mock functions must follow a specific naming convention to be substituted correctly. We are using ld's wrapper functionality to sub these functions. This functionality dictates that in order to wrap a function `foo`, we must have another function `__wrap_foo`. We then pass the linker the option `--wrap=foo`. Anything linked with this flag have its calls to `foo` replaced by `__wrap_foo`.


telemetry/test/telemetry/telemetry.c
```diff
@@ -0,0 +1,102 @@
...
+static void test_telemetry_publish(void)
+{
+    telemetry_packet packet;
+    init_telemetry_queue();
+    TEST_ASSERT_EQUAL_INT(telemetry_publish(packet), true);
+    free_telemetry_queue();
+}
+
+
+K_TEST_MAIN()
```
davesims: @plauche This is _awesome_. 💯 Love the mocking, love the TDD here. 🎉 

I'm wondering if there's a way to dynamically pull all the test files to run in with file globbing, rather than having to declare each test to run in `main`?

ipc/module.json
```diff
@@ -0,0 +1,16 @@
+{
+    "license":"Apache-2.0",
+    "name":"ipc",
+    "version":"0.1.0",
+    "dependencies":{
+      "csp":"kubostech/libcsp"
```
kyleparrott: Are version specifiers necessary for the dependencies?

davesims:
Slam dunk @plauche! ✋ ✖️ 💯 This is fantastic. I've just started reading through and I plan on going over all of it slowly just to learn more about this part of our project -- one great side effect of this approach is the documentation.

🏆 🏆 🏆 🏆 🏆 🏆 🏆 🏆 🏆 🏆 🏆 🏆 

Couple of questions: is it necessary to bring in the whole source of cmocka? If so can we nest it in, say, a `Modules` directory or something under `test`?

plauche:
@davesims Yes it is necessary to bring in the source of cmocka as it is a library which is built with & against our source. Currently cmocka has been brought in as a yotta module in the base source tree so that any module can use it for testing, similar to how the unity library is currently used.

ipc/source/pubsub.c
```diff
@@ -0,0 +1,145 @@
+/*
+ * Copyright (C) 2016 Kubos Corporation
```
marshall: nit: it's 2017 :)

ipc/source/pubsub.c
```diff
@@ -0,0 +1,145 @@
...
+
+    if (csp_listen(*socket, num_connections) != CSP_ERR_NONE)
+    {
+        return false;
+    }
+
+    return true;
+}
+
+bool server_accept(csp_socket_t ** socket, pubsub_conn * conn)
```
marshall: it's not immediately clear why this requires an extra layer of pointer abstraction, you only seem to deref the pointer anyway at line 50, which means `csp_accept` is just getting the the first pointer.

ipc/source/pubsub.c
```diff
@@ -0,0 +1,145 @@
...
+        return true;
+    }
+    else
+    {
+        conn->conn_handle = NULL;
+        return false;
+    }
+}
+
+bool send_csp(pubsub_conn conn, void * data, uint16_t length)
```
marshall: curious why some functions take a `pubsub_conn` by copy, and others (i.e. `server_accept`) take it by pointer?

telemetry/source/telemetry.c
```diff
@@ -13,14 +13,14 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-#ifdef YOTTA_CFG_TELEMETRY
+// #ifdef YOTTA_CFG_TELEMETRY
```
marshall: nit: remove commented code

telemetry/source/telemetry.c
```diff
@@ -29,25 +29,56 @@
...
+    /* Init CSP with address MY_ADDRESS */
+    csp_init(TELEMETRY_CSP_ADDRESS);
+
+    /* Start router task with 500 word stack, OS task priority 1 */
+    csp_route_start_task(500, 1);
+
+    init_telemetry_queue();
+
+    csp_thread_create(telemetry_get_subs, "TELEM_SUBS", 1000, NULL, 0, &telem_sub_handle);
+    csp_thread_create(telemetry_rx_task, "TELEM_RX", 1000, NULL, 0, &telem_rx_handle);
```
marshall: the stack size/number of buffers/size of buffers should all be yotta configurable ala my kubos-shell-example. also, should we maybe create some sort of cap lazy init function that we can safely call from anywhere we need CSP?

telemetry/source/telemetry.c
```diff
@@ -73,11 +103,11 @@ static void telemetry_send(telemetry_packet packet)
 {
     if(packet.source.data_type == TELEMETRY_TYPE_INT)
     {
-        printf("TELEM:%d:%d:%d\r\n", packet.source.source_id, packet.timestamp, packet.data.i);
+        // printf("TELEM:%d:%d:%d\r\n", packet.source.source_id, packet.timestamp, packet.data.i);
```
marshall: nit: more commented out code

telemetry/source/telemetry.c
```diff
@@ -29,25 +29,56 @@
...
+
+    init_telemetry_queue();
+
+    csp_thread_create(telemetry_get_subs, "TELEM_SUBS", 1000, NULL, 0, &telem_sub_handle);
+    csp_thread_create(telemetry_rx_task, "TELEM_RX", 1000, NULL, 0, &telem_rx_handle);
+}
+
+void init_telemetry_queue()
+{
+    packet_queue = csp_queue_create(NUM_MESSAGE_QUEUE, sizeof(telemetry_packet));
```
CatFreed: Why abstract this into its own function?  If it's only going to be one line, why not just use it in the main telemetry_init() function?

ipc/module.json
plauche: Technically no, I believe that without one it will just use whichever version is available. This does raise a good question. Now that everything is in the same repo, therefore pretty tightly versioned together, do we still want to tightly track/enforce individual module versions? @davesims

ipc/source/pubsub.c
plauche: I reworked this a bit to just return back `csp_socket_t*`

ipc/source/pubsub.c
plauche: The functions `server_accept` and `subscriber_connect` both take a `pubsub_conn` by pointer because they modify the structure. The other functions are just reading it.

telemetry/source/telemetry.c
plauche: I had done this at one point so I could init the queue in a test but it is not necessary anymore.

ipc/source/pubsub.c
```diff
@@ -0,0 +1,145 @@
...
+            memcpy(csp_packet->data, data, length);
+            csp_packet->length = length;
+            if (!csp_send(csp_conn, csp_packet, 1000))
+            {
+                csp_buffer_free(csp_packet);
+                return false;
+            }
+            else
+            {
+                return true;
```
plauche: It looks like the csp interface handles it. In this case, since we are just using loopback interfaces, the packet is actually freed by the receiver (since we are in one process). External interfaces (like KISS or I2C) will free the buffer after it is sent over the wire.

plauche:
@marshall re: locking down CSP usage, I definitely agree that we should do that, however I think it should be broken out into an additional task/PR.

ipc/source/pubsub.c
marshall: WRT what we decided in our offline discussion: pubsub functions should be put in the `kprv_` namespace since they are internal, and the difference in consistency won't matter as much if it's not intended to be a public API

PR #30, opened 2017-01-04
[KUBOS-314] Add linux docs

CatFreed:
Migrating linux docs from kubos-linux-build to here so that they can be added to the main docs page.


docs/Linux Overview.md
```diff
@@ -0,0 +1,126 @@
...
+
+Final Distribution:  Custom bootloader.bin to be loaded into whatever persistent storage is available.
+
+### U-Boot
+
+[Wiki](https://en.wikipedia.org/wiki/Das_U-Boot)
+
+[Site Page](http://www.denx.de/wiki/U-Boot)
+
+U-Boot, at the most basic level, is responsible for loading the kernel (zImage) from persistent storage into the SDRAM.  However, it also provides a
```
kyleparrott: Nit: Double spacing before "However,"

CatFreed: Fixed

kyleparrott: I saw this when looking at the raw text but when I looked at the rich diff it only showed up as a single space. There's a couple of other places where there's similar spacing but if it's not showing up when the markdown is rendered I'm not sure how important it is to fix super minor issues like that.

docs/main.md
```diff
@@ -19,10 +21,19 @@ Looking to build an application on KubOS? Check out our [quick start guide](docs
...
+ 
+### KubOS RT
 
  - [MSP430 Launchpad Guide](docs/MSP430-launchpad-guide.md)
 
  - [STM32F4 Discovery Board Guide](docs/STM32F4-discovery-board-guide.md)
+ 
+### KubOS Linux
+ 
+ - [KubOS Linux Overview](docs/Linux%20Overview.md)
```
kyleparrott: I'm not sure if the space in the `Linux Overview` file name is going to cause an issue with it being referenced as `Linux%20Overview.md` - If this works when the docs are generated then :shipit: but otherwise I would consider removing the space from the file name.

docs/main.md
```diff
@@ -19,10 +21,19 @@ Looking to build an application on KubOS? Check out our [quick start guide](docs
...
+ 
+ - [KubOS Linux on iOBC](docs/Linux%20on%20iOBC.md)
```
kyleparrott: Same comment from above with spaces in the file name.

PR #33, opened 2017-01-11
[KUBOS-207] [WIP] Telemetry Storage

cnalley:
:construction: 
Telemetry storage will provide a convenient means to decode telemetry packets into user specified formats and then save the formatted data to various forms of storage. 

**Done**
In the telemetry-storage layer:
1. Added functions to decode the telemetry packet into a CSV log entry and filename with a specific log entry format and naming scheme. An example filename would be “71.csv” where 7 is the source ID of the telemetry packet and 1 is the subsystem address. A log entry would be the telemetry packet time stamp followed by a separating comma and the packets data either in integer or floating point format i.e. “15000,1.5”.
In the disk layer:
1. Decoded telemetry filenames and log entries are logged through klog in it's standard format. Klog has been modified with a special telemetry log level and log handle and config structs to separate regular log files and telemetry logs as desired. Telemetry and regular logs can also be mixed if desired by using the same klog handle.  

**TODO**
Freertos support for klog needs to be added by using fs, fatfs, and syscalls defined in kubos-core in the stm32-hal syscalls stubs for newlib. 
Addition storage formats and storage destinations need to be added to further outline how those should grow in the future and how the configuration will select them.
Further errno values need to be set by klog indicate the success or failure of storage operations and errors handled.
:construction:


plauche:
@cnalley it would be cool to have a parallel PR open to ukub-log-node showing this code in action.

telemetry-storage/test/telemetry-storage/telemetry_storage.c
```diff
@@ -0,0 +1,88 @@
+/*
```
plauche: Hooray for tests :100: 
It would be good to verify the generated file names and log entries, in addition to the return values.

telemetry-storage/source/telemetry_storage.c
```diff
@@ -0,0 +1,181 @@
...
+        printf("Error, decoded log entry or filename are blank \r\n");
+    }
+    
+    #elif (DATA_OUTPUT_FORMAT == FORMAT_TYPE_HEX) 
+    /* Placeholder for hexidecimal format */
+    #else
+    printf("Telemetry storage format type not found\r\n");
+    #endif
+
+    //printf("Log Entry = %s\n", data_buf_ptr);
```
plauche: nit: commented code

telemetry-storage/test/telemetry-storage/telemetry_storage.c
```diff
@@ -15,15 +15,8 @@
  */
 #include <cmocka.h>
 #include <telemetry/telemetry.h>
-#include "telemetry-storage/telemetry_storage.c"
+#include "source/telemetry_storage.c"
```
plauche: Why are we including a source file here?

cnalley: Ah, I meant to ask you about this. What would the protocol / approach moving forward be for unit test of "private" (static) functions?

plauche: Ahhh yes, ok in that case this is fine.

kubos-core/kubos-core/modules/klog.h
```diff
@@ -61,23 +63,35 @@ extern "C" {
...
 void klog_console(unsigned level, const char *logger, const char *format, ...);
-void klog_file(unsigned level, const char *logger, const char *format, ...);
-void klog_cleanup(void);
+void klog_file(klog_handle *handle, klog_config config, unsigned level, const char *logger, const char *format, ...);
+void klog_cleanup(klog_handle *handle);
 
-#define klog_write(level, logger, ...) do { \
-    if (level <= klog_console_level) { \
+#define klog_write(handle, config, level, logger, ...) do { \
+    if (level <= config.klog_console_level) { \
```
plauche: Could we move these config attributes into the `klog_handle` so that you don't have to pass in the config and the handle for every log call?

plauche:
This looks good to me but I will wait for another review before merging.

kubos-core/source/modules/klog/klog.c
```diff
@@ -30,103 +30,97 @@
...
+    
+    handle->log_file = NULL;
+    handle->current_part = -1;
+    handle->current_part_size = 0;
+    
+    _next_log_file(handle);
+    
+    if (handle->log_file == NULL)
+    {
+        return -1;
```
CatFreed: Can a more descriptive return code be used instead? (Like EINVAL)

kubos-core/test/klog.c
```diff
@@ -39,15 +39,11 @@ inline bool _fstat(char *path, int *size) {
 
 void setUp(void)
 {
-    klog_console_level = LOG_NONE;
+    
 }
```
CatFreed: If this function doesn't do anything, it should be removed.

telemetry-storage/source/telemetry_storage.c
```diff
@@ -0,0 +1,189 @@
...
+            /* Store telemetry packets from the telemetry system */
+            telemetry_store(packet);
+        }
+    }
+}
+
+
+void telemetry_storage_init()
+{
+    TELEMETRY_STORE_THREAD;
```
CatFreed: There's no reason for this to be a macro if it's only going to be used here.

telemetry-storage/source/telemetry_storage.c
```diff
@@ -0,0 +1,189 @@
...
+    if (filename_buf_ptr == NULL || file_extension == NULL) 
+    {
+        return 0;
+    }
+
+    len = snprintf(filename_buf_ptr, FILE_NAME_BUFFER_SIZE, "%u%u%s", source_id, address, file_extension);
+
+    if(len < 0 || len >= FILE_NAME_BUFFER_SIZE) 
+    {
+        printf("Filename char limit exceeded. Have %d, need %d + \\0\n", FILE_NAME_BUFFER_SIZE, len);
```
CatFreed: should be klog?

telemetry-storage/telemetry-storage/config.h
```diff
@@ -0,0 +1,85 @@
...
+
+/* Telemetry storage receive task stack depth */
+#ifndef YOTTA_CFG_TELEMETRY_STORAGE_STACK_DEPTH
+#define STORAGE_TASK_STACK_DEPTH 1000
+#else
+#define STORAGE_TASK_STACK_DEPTH YOTTA_CFG_TELEMETRY_STORAGE_STACK_DEPTH
+#endif
+
+/* Telemetry storage receive task priority */
+#ifndef YOTTA_CFG_TELEMETRY_STORAGE_TASK_PRIORITY
```
kyleparrott: I love all the overridable default values 💯

telemetry-storage/test/telemetry-storage/telemetry_storage.c
```diff
@@ -0,0 +1,121 @@
...
+        .source.subsystem_id = 0, .source.data_type = TELEMETRY_TYPE_INT, \
+        .source.source_id = 1};
+        
+    expect_not_value_count(__wrap_klog_init_file, handle->config.file_path, NULL, 2);
+    expect_not_value_count(__wrap_klog_init_file, handle->config.file_path_len, 0, 2);
+    expect_not_value_count(__wrap_klog_init_file, handle->config.part_size, 0, 2);
+    expect_not_value_count(__wrap_klog_init_file, handle->config.max_parts, 0, 2);
+    expect_in_range_count(__wrap_klog_init_file, handle->config.klog_console_level, 0, LOG_ALL+1, 2);
+    expect_in_range_count(__wrap_klog_init_file, handle->config.klog_file_level, 0, LOG_ALL+1, 2);
+    expect_not_value_count(__wrap_klog_init_file, handle->config.klog_file_logging, 0, 2);
```
kyleparrott: nit: it would be nice to group all of the `expect_not_value_count` and `expect_in_range_count` calls

telemetry-storage/source/telemetry_storage.c
plauche: We should eventually turn all error related printfs into error log calls, but the klog module isn't quite ready for that yet.

PR #34, opened 2017-01-11
Initial integration of Slash with KubOS RT

marshall:
This is a big change pulling in the third-party interactive shell library [Slash](https://github.com/kubostech/slash) along with custom changes to get it building for the various STM32F4 and MSP430 KubOS RT targets. 

Synopsis:
Developers can now enable an interactive shell for their KubOS RT application by simply adding the `slash` dependency in their module.json, and starting the `slash_loop` in a FreeRTOS task.

The slash shell is very streamlined and simple, and our integration offers several built-in commands for CSP and Task management, while also giving developers a way to write their own commands with the `slash_command` set of APIs.

Technical info:
Slash was brought in as a squashed git subtree from our fork at [kubostech/slash](https://github.com/kubostech/slash), which itself is a fork of Space Inventor's fork that added support for CSP. There is only 1 commit with slash changes for KubOS RT: It's possible that we could push changes back out of this repo to our fork and upstream the changes, but not necessary.

Slash itself introduced an interesting problem: we needed to force linking of the entire slash static lib at link time otherwise the built-in CSP commands would get dropped. For that reason, I had to make some (mostly minor) changes to kubos-cli so the slash module could export that as a flag to the application using it with a new `module.json` flag called `linkWholeArchive`

There's a few things left to do before this can be merged, but it's safe to go ahead and start reviewing at this point. Checklist:
- [x] write synopsis of changes
- [x] push kubos-shell-example
- [x] review kubostech/slash@bc7f1cb07b1a85e0b8eeb44a224cfa4975d101a2
- [x] review this PR
- [x] review kubostech/kubos-shell-example@3ceeada
- [x] review / merge dependency: kubostech/kubos-cli#7

/cc @kubostech/devs


slash/module.json
```diff
@@ -0,0 +1,23 @@
...
+  "keywords": [],
+  "author": "Kubos",
+  "homepage": "https://www.kubos.co",
+  "license": "Apache-2.0",
+  "lib": "./src",
+  "extraIncludes": ["include"],
+  "linkWholeArchive": true,
+  "targetDependencies": {
+    "kubos-rt": {
+        "csp": "kubostech/libcsp#~1.5.0",
```
kyleparrott: @plauche: Per the conversation Friday, since our source releases are tagged with a top level tag across all modules and targets, are we still using version specifiers in the module.json?

hal/kubos-hal-stm32f4/source/led.c
```diff
@@ -0,0 +1,29 @@
...
+   * @addtogroup LED
+   * @{
+   */
+
+#include "kubos-hal/led.h"
+#include "kubos-hal/gpio.h"
+
+void k_led_init(int led_pin)
+{
+    k_gpio_init(led_pin, K_GPIO_OUTPUT, K_GPIO_PULL_NONE);
```
marshall: yeah, that's fine. I originally put the code in the port-specific implementation in case we need something weird like a GPIO pull up on some boards / MCUs, but we can always reverse it when that is the case

slash/module.json
marshall: LMK if/which tag I need to change to if that's the case

hal/kubos-hal-stm32f4/source/led.c
CatFreed: Also, can the k_led_on and k_led_off functions be grouped in here as well?  I know that it's a little more overhead to have them be their own functions, rather than defines, but from an organizational perspective it would be nice if all the led functions lived in the same place and had the same construction.

hal/kubos-hal-msp430f5529/source/libc/stdlib.c
```diff
@@ -0,0 +1,36 @@
...
+
+#include "kubos-hal-msp430f5529/msp430-stdio.h"
+
+void atexit(void (*fn)(void)) { } // noop stub
+
+void *calloc (size_t nelem, size_t elsize)
+{
+  register void *ptr;
+  if (nelem == 0 || elsize == 0)
+    nelem = elsize = 1;
```
CatFreed: Why not let the request fail if one of the parameters is zero?

slash/README
```diff
@@ -0,0 +1 @@
+Slash!
```
CatFreed: Either put some documentation here about what slash is and how to use it, or remove this doc.

hal/kubos-hal-msp430f5529/source/libc/stdlib.c
marshall: fair enough.. this was copied directly from the msp430-gcc implementation, but that's not a good excuse. need to double check on calloc behavior to see what's expected

slash/README
marshall: yeah, it's useless but it came in from the import of the slash git repository. I'll remove it for us, and at some point we can add real documentation

marshall:
@kyleparrott @CatFreed latest changes should address your comments, LMK if they look good

slash/module.json
kyleparrott: This is a very minor issue and I think we all need to talk about how we handle it across all of our modules/targets at a later time. I think this is good for now.

CatFreed:
I still disagree with the current implementation of calloc.  From a little bit of searching, it seems like the function should return a NULL pointer if the size is zero.

marshall:
@CatFreed ack, well spotted! I just coded/tested/pushed a fix for the behavior. Thanks for keeping me honest :)

marshall:
@CatFreed can you take a look at kubostech/kubos-shell-example@3ceeada as well to make sure it's sane?

tbrowder1:
Looks ok

PR #44, opened 2017-02-02
Kubos build tool

plauche:
This tool gives the kubos developer the ability to build any target/module combination they desire from the root of the kubos directory. It is designed with the future CI environment in mind.

Current functionality:
* Build any module with any target
* Build any module with all targets
* Build all modules for any target
* Build all modules for all targets
* List all available targets
* List all available modules
* List all modules changed in the last commit
* Also adds the `--everything` flag to `kubos-link` which will link all modules/targets into every module under the kubos source tree


tools/build.py
```diff
@@ -0,0 +1,155 @@
...
+
+def main():
+    parser = argparse.ArgumentParser(
+        description='Builds KubOS modules')
+
+    parser.add_argument('--target', metavar='target',
+                        help='Specifies target to build modules for')
+    parser.add_argument('--module', metavar='module',
+                        help='Specifies modules to build')
+    parser.add_argument('--all_targets', action='store_true', default=False,
```
marshall: nit: for consistency, most programs use dashes to separate multi-word long style args, not underscores. can you change that here and the rest of the args below?

FWIW, the dashes will still be converted to underscores in the returned `args` object, so you shouldn't need to change anything else

tools/build.py
```diff
@@ -0,0 +1,155 @@
...
+
+        if echo:
+            print(' '.join(args))
+        try:
+            if save_output:
+                return subprocess.check_output(args, **kwargs)
+            else:
+                return subprocess.check_call(args, stdout=subprocess.PIPE, stderr=subprocess.PIPE, **kwargs)
+        except subprocess.CalledProcessError, e:
+            #print('Error executing command, giving up',file=sys.stderr)
```
CatFreed: nit: remove unneeded code

tools/build.py
```diff
@@ -0,0 +1,155 @@
...
+        return modules
+
+    def list_changed_modules(self):
+        try:
+            git_output = subprocess.check_output(["git", "diff", "--numstat", "HEAD^!"])
+            git_lines = [l for l in git_output.splitlines()]
+            file_paths = [l.split()[2] for l in git_lines]
+            modules = set()
+            for path in file_paths:
+                modules = modules | (self.find_modules(path))
```
marshall: ooh fancy set operations 😻

tools/build.py
```diff
@@ -0,0 +1,155 @@
...
+
+    def build(self, module_name="", target_name=""):
+        #module_list = next((m for m in self.kb.modules() if m.yotta_name() == module_name), None)
+        module = next((m for m in self.kb.modules() if m.yotta_name() == module_name), None)
+        target = next((t for t in self.kb.targets() if t.yotta_name() == target_name), None)
+        if module and target:
+            print('Building [module %s@%s] for [target %s] - ' % (module.yotta_name(), module.path, target_name), end="")
+            self.cmd('kubos', 'target', target_name, cwd=module.path)
+            self.cmd('kubos', 'clean', cwd=module.path)
+            ret = self.cmd('yt', 'build', cwd=module.path)
```
CatFreed: why not kubos build?

tools/build.py
```diff
@@ -0,0 +1,155 @@
...
+        for module in self.kb.modules():
+            print(module.yotta_name())
+
+    def find_modules(self, path):
+        path_list = path.split("/")
+        modules = set()
+        # Pop off file name for first directory
+        path_list.pop()
+        while len(path_list):
+            new_path = "" + "/".join(path_list)
```
marshall: nit: pretty sure the preceding empty string is unnecessary here

tools/build.py
```diff
@@ -0,0 +1,155 @@
...
+            if save_output:
+                return subprocess.check_output(args, **kwargs)
+            else:
+                return subprocess.check_call(args, stdout=subprocess.PIPE, stderr=subprocess.PIPE, **kwargs)
+        except subprocess.CalledProcessError, e:
+            #print('Error executing command, giving up',file=sys.stderr)
+            return 1
+
+    def build(self, module_name="", target_name=""):
+        #module_list = next((m for m in self.kb.modules() if m.yotta_name() == module_name), None)
```
marshall: nit: dead code here and line 32

tools/build.py
```diff
@@ -0,0 +1,155 @@
...
+        if echo:
+            print(' '.join(args))
+        try:
+            if save_output:
+                return subprocess.check_output(args, **kwargs)
+            else:
+                return subprocess.check_call(args, stdout=subprocess.PIPE, stderr=subprocess.PIPE, **kwargs)
+        except subprocess.CalledProcessError, e:
+            #print('Error executing command, giving up',file=sys.stderr)
+            return 1
```
marshall: this code looks familiar, was it pasted from somewhere else / should it be in a library of some kind?

tools/build.py
```diff
@@ -0,0 +1,155 @@
...
+            for m in modules:
+                print(m)
+        except subprocess.CalledProcessError:
+            print("Error getting changed modules")
+
+    def build_all_targets(self, module_name=""):
+        module = next((m for m in self.kb.modules() if m.yotta_name() == module_name), None)
+        if module:
+            for target in self.kb.targets():
+                self.build(module.yotta_name(), target.yotta_name())
```
CatFreed: If the module name doesn't exist, will the "Module * was not found" message be issued for each target in this list?  If so, is there a way to bail if build fails like that?

tools/build.py
```diff
@@ -0,0 +1,155 @@
...
+    
+    def build_all_combinations(self):
+        for target in self.kb.targets():
+            for module in self.kb.modules():
+                self.build(module.yotta_name(), target.yotta_name())
+
+def main():
+    parser = argparse.ArgumentParser(
+        description='Builds KubOS modules')
```
CatFreed: nit: Should be 'Kubos' since the modules aren't an operating system.

tools/build.py
```diff
@@ -0,0 +1,155 @@
...
+                        help='Specifies target to build modules for')
+    parser.add_argument('--module', metavar='module',
+                        help='Specifies modules to build')
+    parser.add_argument('--all_targets', action='store_true', default=False,
+                        help='Builds module for all targets')
+    parser.add_argument('--all_modules', action='store_true', default=False,
+                        help='Builds all modules for target')
+    parser.add_argument('--list_targets', action='store_true', default=False)
+    parser.add_argument('--list_modules', action='store_true', default=False)
+    parser.add_argument('--list_changed_modules', action='store_true', default=False)
```
CatFreed: Worth it to create short versions of some/all of these?

tools/build.py
```diff
@@ -0,0 +1,155 @@
...
+    if args.target and args.module:
+        ret = builder.build(module_name=args.module, target_name=args.target)
+    elif args.module and args.all_targets:
+        builder.build_all_targets(module_name=args.module)
+    elif args.target and args.all_modules:
+        builder.build_all_modules(target_name=args.target)
+    elif args.all_targets and args.all_modules:
+        builder.build_all_combinations()
+    else:
+        ret = -1
```
CatFreed: Will the usage be printed at some point if we hit this case?

tools/build.py
```diff
@@ -0,0 +1,155 @@
...
+        try:
+            git_output = subprocess.check_output(["git", "diff", "--numstat", "HEAD^!"])
+            git_lines = [l for l in git_output.splitlines()]
+            file_paths = [l.split()[2] for l in git_lines]
+            modules = set()
+            for path in file_paths:
+                modules = modules | (self.find_modules(path))
+            
+            if len(modules):
+                print("Modules changed in last commit:")
```
kyleparrott: This only checks between the current and previous commit, correct? Why not diff against master or add an argument for the ref to diff against?

tools/build.py
```diff
@@ -0,0 +1,155 @@
...
+
+    def build_all_modules(self, target_name=""):
+        target = next((t for t in self.kb.targets() if t.yotta_name() == target_name), None)
+        if target:
+            for module in self.kb.modules():
+                self.build(module.yotta_name(), target.yotta_name())
+    
+    def build_all_combinations(self):
+        for target in self.kb.targets():
+            for module in self.kb.modules():
```
kyleparrott: I think it's fine for this iteration but we should add some logic to only build linux and rt applications for their respective target types.

tools/kubos_build.py
```diff
@@ -100,16 +100,18 @@ def find_projects(self):
...
-            relpath = os.path.relpath(path, self.kubos_dir)
-            self.projects.append(Project(name, path, relpath))
+            if "yotta_modules" not in path:
+                path = path.replace("module.json", "").strip()
+                name = path.split("/")[-2]
+                path = os.path.abspath(self.kubos_dir + "/" + path)
+                self.projects.append(Project(name, path))
 
         modules = subprocess.check_output(["find", ".", "-name",
                                         "target.json"], cwd=self.kubos_dir)
```
plauche: I realize it's not the cleanest, bu using the `glob` command doesn't get us the same recursive search and using something like `os.walk` takes a bit longer than the Linux find command.

tools/build.py
plauche: Currently the `kubos build` command does not return a meaningful return code, but `yotta build` does.

plauche: Yup that is it

tools/build.py
plauche: Yup currently it is just the previous commit. I've modified it to compare against the previous commit by default, but you could also pass in a custom diff path.

tools/build.py
plauche: Yup, we need to come up with a good way of identifying which modules should be built which way (or if they don't need to be build).

tools/build.py
```diff
@@ -0,0 +1,151 @@
...
+        self.kb = KubosBuild()
+        self.modules = self.kb.modules()
+        self.targets = self.kb.targets()
+
+    def build(self, module_name="", target_name=""):
+        module = next((m for m in self.kb.modules() if m.yotta_name() == module_name), None)
+        target = next((t for t in self.kb.targets() if t.yotta_name() == target_name), None)
+        if module and target:
+            print('Building [module %s@%s] for [target %s] - ' % (module.yotta_name(), module.path, target_name), end="")
+            utils.cmd('kubos', 'target', target_name, cwd=module.path, echo=False, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```
kyleparrott: One assumption that this makes is that the required target(s) are linked into the current module that is being built. A solution for this is to link all of the modules and targets like `utils.cmd('kubos', 'link', '--all', cwd=module.path)` before setting the target.

plauche: Yup that is correct. We already have the kubos_link tool in place, could we just use them one after the other?

kyleparrott: In the slightly dangerous instance when we would run this tool with `--all-modules --all-targets` would we have to run the link tool in some sort of loop to link everything to all the modules before running this script? Or does the link tool already have an option to "Link everything to everything" for lack of a better term.

I see CI running some sort of command(s) that will build every module for every appropriate target which is why I was looking at this case.

plauche: This PR also adds a `--modules` option to the `kubos_link` tool which will link all modules into each module under the kubos source tree.

kyleparrott: Awesome - I didn't see that. That's perfect.

PR #45, opened 2017-02-03
Created initial v0.2 changelog list

CatFreed:
Created a high-level list of everything we've done since October...

**NOTE:** This does not include CIStack or the Dashboard.  Wasn't sure a) if they should be included here and b) how they should be listed

@kubostech/devs


marshall:
Ship it!

PR #53, opened 2017-02-08
[KUBOS-350][WIP] Telemetry for multiple processes

plauche:

PR #55, opened 2017-02-13
KUBOS-388 update for the command and control telemetry daemon and client.

kyleparrott:
The goal of this is to build out an example command and control daemon and client. The remaining tasks to do are: 

- [X] Send a command from a client
- [X] Receive a command from the client
- [X] Parse the action, command and remaining arguments from the command
- [x] Make argument handling more dynamic
- [X] Load and execute the command and associated action function from a dynamic library
- [x] Send the output and other metadata about the command execution to the client
- [x] Solidify the input and output data structures
- [x] Capture stdout/stderr for the commands
- [x] Send the stdout/stderr to the client.
- [x] Misc code cleanup
- [x] Clean up argp usage


plauche:
Is this one PR's worth of work or should we break this up into multiple?

kyleparrott:
That is a good question. I'm not sure how it would be split into functional individual components. I'm open to any suggestions on that though :)

kyleparrott:
I have tried to narrow the scope of this PR to the basic command and control functionality. While much of this code is still in need of lots of clean up the basic structure should be mostly implemented, with the exception of sending command output back to the client.

I will continue to clean this up and implement more functionality but I think it's in a place for a high level look over to see if we like this structure and start to figure out what parts of the flow we want to change.

cc/ @kubostech/devs

kyleparrott:
Since this PR is still not complete and is dependent on some of the changes from #53  I don't quite think a line level review is necessary right now. I am looking for a higher level review on the overall structure and implementation of this framework. I am open to all feedback on this PR 😄 

cc/ @kubostech/devs

cmd-control-daemon/source/command.c
```diff
@@ -0,0 +1,130 @@
...
+        if (conn) {
+            packet = csp_read(conn, 0);
+            if (packet)
+            {
+                if (!parse_command_cbor(packet, command)){
+                    fprintf(stderr, "There was an error parsing the command packet\n");
+                    return false;
+                }
+            }
+            csp_buffer_free(packet);
```
CatFreed: potentially freeing null pointer

cmd-control-daemon/source/command.c
```diff
@@ -0,0 +1,130 @@
...
+            csp_close(conn);
+            return true;
```
CatFreed: not returning false if packet's null?

cmd-control-daemon/source/main.c
```diff
@@ -0,0 +1,194 @@
...
+    cnc_response_packet response;
+    cnc_command_wrapper wrapper;
+
+    wrapper.command_packet = &command;
+    csp_init_things(my_address);
+    sock = csp_socket(CSP_SO_NONE);
+    csp_bind(sock, PORT);
+    csp_listen(sock, 5);
+
+    while (1) {
```
CatFreed: Do we want a way to exit the daemon gracefully? So, while(!exit) or something instead?

cmd-control-daemon/source/main.c
```diff
@@ -0,0 +1,194 @@
...
+        zero_vars(command_str, &command, &response);
+        get_command(sock, command_str);
+        parse(command_str, &wrapper);
+        run_command(&wrapper, &response);
+        send_response(&response);
```
CatFreed: If there's a problem with any of these steps, you should probably bail on the command and either do some error handling or just continue to the next loop.

cmd-control-daemon/source/parser.c
```diff
@@ -0,0 +1,153 @@
...
+#include <string.h>
+
+#include "command-and-control/types.h"
+#include "cmd-control-daemon/daemon.h"
+
+//Action string hash values
+#define EXEC    6385204650
+#define HELP    6385292014
+#define OUTPUT  6953876217206
+#define STATUS  6954030894409
```
CatFreed: Where did these hash values come from?

cmd-control-daemon/source/parser.c
```diff
@@ -0,0 +1,153 @@
...
+static int parse_opt (int key, char *arg, struct argp_state *state)
+{
+    cnc_command_wrapper *arguments = state->input;
+    int idx;
+    switch (key)
+    {
+        case ARGP_KEY_ARG:
+            {
+                printf("ARG: %s\n", arg);
+                switch(arguments->command_packet->arg_count++)
```
CatFreed: I'm not quite following the code.  What's the purpose of this switch statement?

cmd-control-daemon/source/parser.c
```diff
@@ -0,0 +1,153 @@
...
+            {
+                //Do some validation to make sure we have a minimum number of arguments
+                if (strlen(arguments->command_packet->cmd_name) == 0 || !arguments->command_packet->action)
+                {
+                    send_usage_error(arguments);
+                    fprintf(stderr, "received incorrect command or action argument\n");
+                }
+                arguments->command_packet->arg_count = arguments->command_packet->arg_count - 2;
+                printf("End of parsing\n");
+                return -1;
```
CatFreed: -1 normally indicates a bad return code, but it seems like reaching the end of parsing is a good thing.  Maybe a different code?  (Except if the previous 'incorrect command' statement gets hit)

cmd-control-daemon/source/parser.c
```diff
@@ -0,0 +1,153 @@
...
+            }
+            break;
+    }
+    return 0;
+}
+
+
+int get_num_args(char* string){
+    int count = 0, i = 0;
```
CatFreed: Why not just start count and i at 1, since you're going to immediately increment them anyways?

cmd-control-client/source/main.c
```diff
@@ -0,0 +1,204 @@
...
+#include <csp/csp_interface.h>
+
+#include "command-and-control/types.h"
+#include "tinycbor/cbor.h"
+
+#define PORT        10
+#define BUF_SIZE    MTU
+
+pthread_t rx_thread, my_thread;
+int rx_channel, tx_channel;
```
kyleparrott: This will switch to using a unix socket so all of the named pipe specifics will go away in the not so distant future. I mentioned in the [daemon code](https://github.com/kubostech/kubos/blob/command-and-control-prototype/cmd-control-daemon/source/main.c#L20) but forgot to include the same comment here, explaining that the first ~89 lines of this file are basically the csp fifo/named pipe example. 

This code will not be used and will be replaced with the socket implementation once #53 is merged.

cmd-control-daemon/source/parser.c
kyleparrott: It's for assigning the argument value depending on its order. 

Argument 1  -> the action from the command interface to perform on the library (execute, status, help, etc..)
Argument 2  -> The library/group to perform the action on ( group "foo" would load "libfoo.so")
Argument 3 - N -> The trailing arguments Are all passed to the library command to do with as it pleases.

I don't like this because it's not using the actually argp for anything more than a looping mechanism, which defeats the purpose of using the argument parsing library in the first place. This will get replaced with a better Argp usage in the very near future.

kyleparrott:
Alright, This should be mostly good to look at closely. There are a couple of issues:

* This still uses named pipes. I don't see this being merged before #53 so I propose that we mostly ignore that set up code (it's going to be thrown away soon anyway) in the client and daemon as it's the CSP fifo example code.

* There's still some print statements in there. Logging isn't really a thing yet. I see these print statements being replaced with logging calls once we have implemented logging.

davesims:
👍 ✨ Great work @kyleparrott. I'll get 👀 on it this afternoon.

davesims:
A couple of initial comments for posterity. We mentioned this earlier this week, but I'll reiterate here: let's get better at breaking these PRs up. One good discipline is to set a TODO on a PR and stick to less than 5 line items if possible, 3-4 is ideal.

As to code review granularity -- in general I'd encourage reviews that treat code as the final word and not a rough draft. It's OK to be picky about things like naming and code conventions. Where there can be reasonable disagreement (ternaries? brackets for all if statements?) let's move towards identifying a single coding standard and eventually automate linting against that standard in CI.

Me, I just ran rubocop (the standard ruby linter) against my code and it made me want to do the Charlie Brown sad walk.



Also, check out Phil Haack's blog on code review, it's worth reading a few times. 

http://haacked.com/archive/2013/10/28/code-review-like-you-mean-it.aspx/

davesims:
The above :point_up: having been said, I'd say this PR is actually not too bad, size-wise. Could have been split once I think, but that's all.

cmd-control-client/source/main.c
```diff
@@ -0,0 +1,260 @@
...
+    int server_address = 1;
+    csp_conn_t *conn;
+    csp_packet_t *packet;
+
+    while(packet = csp_buffer_get(length)) {
+        memcpy(packet->data, data, length);
+        packet->length = length;
+
+        conn = csp_connect(CSP_PRIO_NORM, server_address, PORT, 1000, CSP_O_NONE);
+        send_packet(conn, packet);
```
CatFreed: Worth it to check for and pass back a bad return code here?

cmd-control-client/source/main.c
```diff
@@ -0,0 +1,260 @@
...
+    if (err || cbor_value_get_int(&element, &message_type))
+        return false;
+
+    switch (message_type) {
+        case RESPONSE_TYPE_COMMAND_RESULT:
+            return parse_command_result(&parser, &map);
+        case RESPONSE_TYPE_PROCESSING_ERROR:
+            return parse_processing_error(&parser, &map);
+        default:
+            fprintf(stderr, "Recevied unknown message type: %i\n", message_type);
```
CatFreed: nit: spelling

cmd-control-client/source/main.c
```diff
@@ -0,0 +1,260 @@
...
+}
+
+
+void send_msg(uint8_t* data, size_t length)
+{
+    int server_address = 1;
+    csp_conn_t *conn;
+    csp_packet_t *packet;
+
+    while(packet = csp_buffer_get(length)) {
```
CatFreed: this should be an if statement instead of a while

cmd-control-client/source/main.c
```diff
@@ -0,0 +1,260 @@
...
+{
+    csp_socket_t *sock;
+    csp_conn_t *conn;
+    csp_packet_t *packet;
+
+    sock = csp_socket(CSP_SO_NONE);
+    csp_bind(sock, PORT);
+    csp_listen(sock, 5);
+
+    while (conn = csp_accept(sock, 1000)) {
```
CatFreed: This should be an if statement instead of a while

cmd-control-client/source/main.c
```diff
@@ -0,0 +1,260 @@
...
+            return;
+        }
+    }
+}
+
+
+int main(int argc, char **argv)
+{
+    //terrible name for starting all the tedious csp stuff
+    if (!init())
```
CatFreed: nit: In C, all variables are normally declared at the top of the function, before any work is done

cmd-control-client/source/main.c
```diff
@@ -0,0 +1,260 @@
...
+    uint8_t data[BUF_SIZE] = {0};
+
+    CborEncoder encoder, container;
+    CborError err;
+
+    int i;
+    for (i = 1; i < argc; i++)
+    {
+        strcat(args, argv[i]);
+        if (i != argc -1) //Skip the final separator
```
CatFreed: nit: remove the space between 'argc' and '-1'

cmd-control-client/source/main.c
```diff
@@ -0,0 +1,260 @@
...
+
+    char* separator = " ";
+    char args[BUF_SIZE] = {0};
+    uint8_t data[BUF_SIZE] = {0};
+
+    CborEncoder encoder, container;
+    CborError err;
+
+    int i;
+    for (i = 1; i < argc; i++)
```
CatFreed: Unless you're sure there's no way to enter args with more than BUF_SIZE characters, you should do length checking during this loop

cmd-control-daemon/README.md
```diff
@@ -0,0 +1,6 @@
+# KubOS Linux Example App
```
CatFreed: Update this

cmd-control-daemon/source/command.c
```diff
@@ -0,0 +1,176 @@
...
+    {
+        wrapper->err = true;
+        snprintf(wrapper->output, sizeof(wrapper->output) - 1, "Unable to open lib: %s\n", dlerror());
+        return false;
+    }
+
+    switch (wrapper->command_packet->action)
+    {
+        case execute:
+            *func = dlsym(*handle, "execute");
```
CatFreed: typo?  '*func' should be 'func' like the other commands

cmd-control-daemon/source/command.c
```diff
@@ -0,0 +1,176 @@
...
+            func = dlsym(*handle, "help");
+            break;
+        default:
+            snprintf(wrapper->output, sizeof(wrapper->output) - 1, "Unable to open lib: %s\n", dlerror());
+            return false;
+    }
+
+    if (func == NULL)
+    {
+        snprintf(wrapper->output, sizeof(wrapper->output) - 1, "The requested Symbol doesn't exist\n");
```
CatFreed: Nit: lowercase symbol

cmd-control-daemon/source/command.c
```diff
@@ -0,0 +1,176 @@
...
+    freopen("/dev/null", "a", stdout);
+    dup2(original_stdout, STDOUT_FILENO); //restore the previous state of stdout
+    setbuf(stdout, NULL);
+
+    //Calculate the runtime
+    wrapper->response_packet->execution_time = (double)(finish_time - start_time) / (CLOCKS_PER_SEC/1000); //execution time in milliseconds
+    /*printf("Return code: %i exection time %f\n", wrapper->response_packet->return_code, wrapper->response_packet->execution_time);*/
+
+    //Unload the library
+    func = NULL;
```
CatFreed: This doesn't do anything useful, since func is a passed-in value rather than a pointer?

cmd-control-daemon/source/main.c
```diff
@@ -0,0 +1,181 @@
...
+        packet = csp_buffer_get(BUF_SIZE);
+        if (packet) {
+            memcpy(packet->data, data, data_len);
+            packet->length = data_len;
+
+            conn = csp_connect(CSP_PRIO_NORM, client_address, PORT, 1000, CSP_O_NONE);
+            send_packet(conn, packet);
+            csp_buffer_free(packet);
+            csp_close(conn);
+            return true;
```
CatFreed: This shouldn't return true if send_packet failed

cmd-control-daemon/source/main.c
```diff
@@ -0,0 +1,181 @@
...
+
+
+int main(int argc, char **argv)
+{
+    int my_address = 1;
+    csp_socket_t *sock;
+    char command_str[75];
+    cnc_command_packet command;
+    cnc_response_packet response;
+    //The wrapper keeps track of a command input, it's result and
```
CatFreed: nit: its

cmd-control-daemon/source/parser.c
```diff
@@ -0,0 +1,144 @@
...
+            }
+            break;
+        case ARGP_KEY_END:
+            if (strlen(arguments->command_packet->cmd_name) == 0) //TODO: Effectively validate the action
+            {
+                arguments->err = true;
+                snprintf(arguments->output, sizeof(arguments->output) - 1, "received incorrect command or action argument\n"); //Would be helpful to give the help message here..
+                send_result(arguments);
+            }
+            arguments->command_packet->arg_count = arguments->command_packet->arg_count - 2;
```
CatFreed: nit: add break;

cmd-control-daemon/source/parser.c
```diff
@@ -0,0 +1,144 @@
...
+        if (string[i] == ' ') count++;
+    }
+    return count + 1;
+}
+
+
+static char args_doc[] = "Action Group-Name [following args]";
+static char doc[] = "CNC Daemon - Execute commands through the Kubos command and control framework";
+
+static struct argp argp = { options, parse_opt, args_doc, doc};
```
CatFreed: nit: move these declarations to below the defines, but above the first function

cmd-control-sample-so/source/main.c
```diff
@@ -0,0 +1,21 @@
+#include <stdio.h>
+
+int execute(int argc, char **argv) {
+    //printf("Calledexecute called with argc: %i, argv: %s\n", argc, *argv);
```
CatFreed: nit: remove unused lines

command-and-control/command-and-control/types.h
```diff
@@ -0,0 +1,90 @@
...
+
+#ifdef YOTTA_CFG_CNC_CMD_PACKET_ARG_LEN
+#define CMD_PACKET_ARG_LEN        YOTTA_CFG_CNC_CMD_PACKET_ARG_LEN
+#else
+#define CMD_PACKET_ARG_LEN        DEFAULT_STR_LEN
+#endif
+
+//The size of all the members of the command packet, except the output field
+//The packet must fit into the CSP MTU or bad things will happen
+#define CMD_PACKET_MEMBER_SIZE sizeof(int) + sizeof(cnc_action) + sizeof(char) * CMD_PACKET_CMD_NAME_LEN
```
CatFreed: nit: char is always 1 byte, so sizeof(char) is unnecessary

cmd-control-daemon/cmd-control-daemon/daemon.h
```diff
@@ -0,0 +1,29 @@
+#ifndef DAEMON_H
```
davesims: BTW, Great work defining a clean interface @kyleparrott. ✨ Overall a great job with organization & boundaries.

cmd-control-daemon/source/command.c
```diff
@@ -0,0 +1,176 @@
...
+    clock_t finish_time = clock();
+
+    //Redirect stdout back to the terminal.
+    freopen("/dev/null", "a", stdout);
+    dup2(original_stdout, STDOUT_FILENO); //restore the previous state of stdout
+    setbuf(stdout, NULL);
+
+    //Calculate the runtime
+    wrapper->response_packet->execution_time = (double)(finish_time - start_time) / (CLOCKS_PER_SEC/1000); //execution time in milliseconds
+    /*printf("Return code: %i exection time %f\n", wrapper->response_packet->return_code, wrapper->response_packet->execution_time);*/
```
davesims: Do we want to uncomment this, or make it a function of a verbose level option/logging level?

cmd-control-client/source/main.c
```diff
@@ -0,0 +1,260 @@
...
+    csp_route_start_task(0, 0);
+    return true;
+}
+
+//Where the magic happens - Bascially ignore everything above this line - The initialization is going to change a lot.
+
+int send_packet(csp_conn_t* conn, csp_packet_t* packet)
+{
+    if (!conn || !csp_send(conn, packet, 1000))
+        return -1;
```
davesims: brackets. :trollface:

cmd-control-daemon/source/main.c
davesims: Agreed. Should it return the value of `send_packet`?

command-and-control/command-and-control/types.h
```diff
@@ -0,0 +1,90 @@
...
+#define RES_PACKET_STDOUT_LEN        YOTTA_CFG_CNC_CMD_PACKET_ARG_LEN
+#else
+#define RES_PACKET_STDOUT_LEN     MTU - CMD_PACKET_MEMBER_SIZE
+#endif
+
+#define RESPONSE_TYPE_COMMAND_RESULT 0
+#define RESPONSE_TYPE_PROCESSING_ERROR 1
+
+
+typedef enum
```
davesims: Also this is a possibility: https://github.com/google/styleguide/tree/gh-pages/cpplint

/cc @marshall @plauche @CatFreed for 💭

kyleparrott:
Another more fundamental question I have about this PR is if the client or daemon need to be in their own repo(s). I remember a discussion a while back about keeping executables out of the Kubos repo..

cmd-control-client/source/main.c
```diff
@@ -0,0 +1,260 @@
...
+            return parse_command_result(&parser, &map);
+        case RESPONSE_TYPE_PROCESSING_ERROR:
+            return parse_processing_error(&parser, &map);
+        default:
+            fprintf(stderr, "Recevied unknown message type: %i\n", message_type);
+            return false;
+   }
+}
+
+bool parse_command_result( CborParser * parser, CborValue * map)
```
kyleparrott: I think these message parsing and the encoding functions should all move to the higher level command and control module and out of the client code. Ideally moving forward there will be numerous clients that will all use these. Thoughts?

command-and-control/command-and-control/types.h
plauche: I have played some with https://clang.llvm.org/docs/ClangFormat.html, it is pretty flexible and will actually reformat your file in-place. It looks like cpplint only reports style violations.

cmd-control-daemon/source/main.c
```diff
@@ -0,0 +1,181 @@
...
+#include <stdint.h>
+#include <stdio.h>
+#include <string.h>
+#include <unistd.h>
+
+#include "command-and-control/types.h"
+#include "cmd-control-daemon/daemon.h"
+#include "tinycbor/cbor.h"
+
+#define PORT        10
```
kyleparrott: That's the csp port, not TCP/UDP. CSP only supports ports 0-255 :)

cmd-control-daemon/source/command.c
kyleparrott: It was a typo in the opposite way. I was testing on the exec Command exclusively for a while and missed changing the others.

cmd-control-daemon/source/main.c
```diff
@@ -0,0 +1,181 @@
...
+    .name = "fifo",
+    .nexthop = csp_fifo_tx,
+    .mtu = MTU,
+};
+
+
+int csp_fifo_tx(csp_iface_t *ifc, csp_packet_t *packet, uint32_t timeout) {
+    /* Write packet to fifo */
+    if (write(tx_channel, &packet->length, packet->length + sizeof(uint32_t) + sizeof(uint16_t)) < 0)
+        printf("Failed to write frame\r\n");
```
kyleparrott: I left the brackets, such rigid path definitions and general ugliness in this block because it's going to be deleted well before this PR is merged.

This block is getting deleted in favor of the socket interface, as the comments on lines 20, 92, and a few of the entries in the conversation on this PR state. 

I expedited this PR for review by leaving this block in so that way we could get moving on review of all the other parts of this PR since this is slightly longer/more complex than typical PRs. 

I was not sure how long review would take. I didn't want to delay the next release by waiting on review of this. I will be more clear in the future.

cmd-control-daemon/source/main.c
davesims: ...right. Because there is no TCP/IP stack. That's the whole point of CSP. I'm not sure why anyone would assume otherwise. You pass the test @kyleparrott. 🤕

cmd-control-client/module.json
```diff
@@ -3,15 +3,14 @@
     "license":"Apache-2.0",
     "name":"cmd-control-client",
     "repository":{
-        "url":"git://<repository_url>",
+        "url":"git://github.com/kubostech/kubos",
```
davesims: 👍

cmd-control-daemon/source/command.c
```diff
@@ -66,13 +72,13 @@ bool load_command(cnc_command_wrapper * wrapper, void ** handle, lib_function *
 {
     int return_code;
     char so_path[SO_PATH_LENGTH];
-    char * home_dir = "/home/vagrant/lib%s.so"; //Just a dev path for now.
 
     // so_len - the format specifier length (-2) + the null character (+1) leading to the -1
-    int so_len = strlen(home_dir) + strlen(wrapper->command_packet->cmd_name) - 1;
-    snprintf(so_path, so_len, home_dir, wrapper->command_packet->cmd_name);
+    int so_len = strlen(MODULE_REGISTRY_DIR) + strlen(wrapper->command_packet->cmd_name) - 1;
```
davesims: 👍

command-and-control/command-and-control/types.h
```diff
@@ -0,0 +1,94 @@
...
+
+
+// Used inside the daemon to track and provide error messages back to the client
+typedef struct
+{
+    cnc_command_packet  * command_packet;
+    cnc_response_packet * response_packet;
+    bool err;
+    char output[RES_PACKET_STDOUT_LEN];
+} cnc_command_wrapper;
```
davesims: I'm of the opinion that going forward struct & type names should be camel-cased, i.e. `CNCComandWrapper`. This makes it easier to :eyes: method signatures & other code to distinguish types from normal vars.

https://google.github.io/styleguide/cppguide.html#Type_Names

Also, is this a 'wrapper' or a `CNCCommand`? I think I prefer the latter.

command-and-control/source/cnc.c
```diff
@@ -0,0 +1 @@
+
```
davesims: ❓

command-and-control/command-and-control/types.h
davesims: @plauche yeah I like it. Looks like you can also customize it to follow a given standard.

cmd-control-client/source/main.c
CatFreed: Seems logical to me. I could also see waiting until another client is actually implemented and then pulling out all common code, since then you'll _know_ what's common, rather than just guessing what will be.

cmd-control-client/source/main.c
```diff
@@ -0,0 +1,299 @@
...
+{
+    csp_socket_t *sock;
+    csp_conn_t *conn;
+    csp_packet_t *packet;
+
+    sock = csp_socket(CSP_SO_NONE);
+    csp_bind(sock, PORT);
+    csp_listen(sock, 5);
+
+    while (conn = csp_accept(sock, 1000))
```
CatFreed: nit: unnecessary while?

cmd-control-client/source/main.c
```diff
@@ -0,0 +1,204 @@
...
+        fprintf(stderr, "There was an error creating the map. CBOR Error code: %i\n", err);
+        return 0;
+    }
+    err = cbor_encode_text_stringz(&container, "ARGS");
+    if (err || cbor_encode_text_stringz(&container, args)) {
+        fprintf(stderr, "There was an error encoding the commands into the CBOR message\n");
+    }
+
+    send_msg(data, BUF_SIZE);
+    get_response();
```
CatFreed: @kyleparrott are you planning on adding more error reporting in a future PR?

cmd-control-daemon/cmd-control-daemon/daemon.h
CatFreed: nit: copyright in all the new files

cmd-control-client/source/main.c
kyleparrott: I think that is a good idea. I feel this PR is a first pass at this system. I think there's about another PR or twos worth of logging, error handling & recovery and general error checking for sometime in the future.

CatFreed: cool, sounds good

kyleparrott:
After chatting with @plauche - In our next release (not tomorrow) we should move the argument parsing into the client and implement a slightly more complex CBOR format input message. This would eliminate parsing in the daemon and simplify it's duties significantly.

cmd-control-daemon/source/main.c
```diff
@@ -0,0 +1,189 @@
...
+    {
+        conn = csp_accept(sock, 1000);
+        if (conn)
+        {
+            packet = csp_read(conn, 0);
+            if (packet)
+            {
+                if (!parse_command_cbor(packet, command))
+                {
+                    fprintf(stderr, "There was an error parsing the command packet\n");
```
CatFreed: need to free the packet and close the connection here (unless i'm missing some later processing that'll take care of it)

cmd-control-daemon/source/main.c
```diff
@@ -0,0 +1,191 @@
...
+    csp_packet_t *packet;
+
+    if(packet = csp_buffer_get(BUF_SIZE))
+    {
+        memcpy(packet->data, data, data_len);
+        packet->length = data_len;
+
+        conn = csp_connect(CSP_PRIO_NORM, CLI_CLIENT_ADDRESS, CSP_PORT, 1000, CSP_O_NONE);
+        if (!send_packet(conn, packet))
+        {
```
CatFreed: Free packet + close conn

kyleparrott:
@davesims @plauche : Friendly request for 👀

cmd-control-daemon/source/message.c
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    return finish_encode_response_and_send(data, encoder, container);
+}
+
+
+bool finish_encode_response_and_send(uint8_t * data, CborEncoder *encoder, CborEncoder * container)
+{
+    cbor_encoder_close_container(encoder, container);
+    size_t data_len = cbor_encoder_get_buffer_size(encoder, data);
```
plauche: I'm not so sure about this tight interleaving of send-related (`send_result`, `send_buffer`) and encoding-relating functions. This will make it difficult to test these encode functions in a standalone fashion.

tools/utils.py
```diff
@@ -0,0 +1,20 @@
+import subprocess
```
plauche: These changes are already in master, might want to rebase.

cmd-control-client/source/main.c
```diff
@@ -0,0 +1,267 @@
...
+
+    CborValue element;
+    CborError err;
+
+    err = cbor_value_map_find_value(map, "RETURN_CODE", &element);
+    if (err || cbor_value_get_simple_type(&element, &return_code))
+    {
+        return false;
+    }
+    printf("Return Code: %i\n", return_code);
```
kyleparrott: Would you be alright with decoupling the parsing and processing in the next iteration? There's not a lot else to do with them other than print (no logging, etc) at the moment.

cmd-control-daemon/source/message.c
kyleparrott: Would this also be a goal of the next iteration? I think adding unit tests would be an absolute must have for the next version as well.

plauche: Yup I think this would be for the next iteration.

cmd-control-client/source/main.c
plauche: Yup that will work.

PR #57, opened 2017-02-28
[KUBOS-325] Adding detection and routing for flashing KubOS Linux upgrade packages

CatFreed:
Updating the ISIS target's flash script to check for the *.itb file extension.  If found, it will load the file into the new /upgrade partition and set the new kubos_updatefile environment variable with the new file name.

Also, migrating user applications from /usr/bin into /home/usr/bin so that they will live in the new userspace partition, rather than the rootfs.

### DIFF
--- a/targets/target-kubos-linux-isis-gcc/minicom/flash.sh
+++ b/targets/target-kubos-linux-isis-gcc/minicom/flash.sh
@@ -28,6 +28,7 @@ start=$(date +%s)
 this_dir=$(cd "`dirname "$0"`"; pwd)
 program=$1
 name=$(basename $1)
+is_upgrade=0
 
 password=$(cat yotta_config.json | python -c 'import sys,json; x=json.load(sys.stdin); print x["system"]["password"]')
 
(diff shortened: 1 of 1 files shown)


targets/target-kubos-linux-isis-gcc/minicom/flash.sh
```diff
@@ -65,17 +73,18 @@ expect {
     "~ #" break
     timeout 5 goto end
 }
-send "cd /usr/bin"
+send "cd $path"
 send "rm $name"
 send "rz -bZ"
 ! sz -b --zmodem $1
+if $is_upgrade = 1 send "fw_setenv kubos_updatefile $name"
```
plauche: I think this needs to be
`if [[ $is_upgrade = 1 ]]; then send "fw_setenv kubos_updatefile $name"; fi`

plauche: Nvm I forgot this was inside of a block of minicom script :)

PR #58, opened 2017-03-02
[KUBOS-350] CSP Modifications

plauche:
* Switched tcp socket interface to use SCTP for more reliable message delivery
* Introduced cbor serialization for tcp socket communication
* Added some termination/cleanup functions
* Modified csp build structure to allow for yotta building of tests
* Added unit test and integration tests to exercise tcp socket interface

PR #59, opened 2017-03-02
[KUBOS-350] IPC Changes

plauche:
Changes to the ipc layer in support for Linux Telemetry
- [x] Added new functions specific to opening/closing socket connections
- [x] Added integration tests for socket functionality
- [ ] Migrate basic csp init/cleanup functions to here

Branched from and dependent on #58


ipc/source/pubsub.c
```diff
@@ -43,17 +46,44 @@ csp_socket_t * kprv_server_setup(uint8_t port, uint8_t num_connections)
...
+    }
+    if ((csp_conn = csp_accept(socket, 1000)) != NULL)
+    {
+        conn->conn_handle = csp_conn;
+        return true;
     }
     return false;
 }
 
+bool kprv_server_socket_accept(csp_socket_t * socket, pubsub_conn * conn)
```
CatFreed: What's the intended difference between using kprv_server_socket_accept and kprv_server_accept?

plauche: The `kprv_server_accept` is just accepting a connection over a standard csp socket. The `kprv_server_socket_accept` creates a TCP socket connection and then attempts to accept a csp socket over that.

.clang-format
```diff
@@ -0,0 +1,90 @@
...
+SpaceBeforeAssignmentOperators: true
+SpaceBeforeParens: ControlStatements
+SpaceInEmptyParentheses: false
+SpacesBeforeTrailingComments: 1
+SpacesInAngles:  false
+SpacesInContainerLiterals: true
+SpacesInCStyleCastParentheses: false
+SpacesInParentheses: false
+SpacesInSquareBrackets: false
+Standard:        Cpp03
```
kyleparrott: Is the mismatch in the linter cpp03 and the c99 compilation standard we use for several targets significant?

ipc/source/pubsub.c
CatFreed: Cool, ok :+1:

PR #62, opened 2017-03-06
[KUBOS-313] KubOS Linux Non-application File Transfer

CatFreed:
Added code to the iOBC flash script to allow us to detect and transfer non-application files.

I'm operating on the assumption that if the file requested for transfer does not match the project name, then it must not be the application binary. If so, then it will be transferred to the non-app directory location.

This destination can be customized by the user with a new config.json parameter "destDir". (I kind of dislike the variable name. Alternate suggestions would be appreciated)

### DIFF
--- a/targets/target-kubos-linux-isis-gcc/minicom/flash.sh
+++ b/targets/target-kubos-linux-isis-gcc/minicom/flash.sh
@@ -26,6 +26,8 @@ spinner() {
 start=$(date +%s)
 
 this_dir=$(cd "`dirname "$0"`"; pwd)
+project=$(cd "../.."; pwd)
+project=$(basename $project)
 program=$1
 name=$(basename $1)
 is_upgrade=0
@@ -55,6 +57,8 @@ fi
 if [[ "$name" = *.itb ]]; then
     path="/upgrade"
     is_upgrade=1
+else if [[ "$name" != "$project" ]]; then
+    path="/home/usr/local/bin"
 else
     path="/home/usr/bin"
 fi
--- a/targets/target-kubos-linux-isis-gcc/minicom/flash.sh
+++ b/targets/target-kubos-linux-isis-gcc/minicom/flash.sh
@@ -16,7 +16,7 @@
 
 progress() {
     while sleep 1; do
-        line=$(tac flash.log | grep -m 1 "Bytes Sent")
+        line=$(cat flash.log | grep "Bytes Sent")
         printf "$line\r"
     done   
 }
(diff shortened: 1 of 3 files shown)


targets/target-kubos-linux-isis-gcc/minicom/flash.sh
```diff
@@ -26,11 +26,14 @@ spinner() {
 start=$(date +%s)
 
 this_dir=$(cd "`dirname "$0"`"; pwd)
+project=$(cd "../.."; pwd)
```
marshall: shouldn't the project name come from `yotta_config.json` or `module.json`?

targets/target-kubos-linux-isis-gcc/target.json
```diff
@@ -30,6 +30,7 @@
       "arm": {}
     },
     "system": {
+      "destDir": "/home/usr/local/bin",
```
marshall: is this the path we're using in the Linux image? never seen an alternate `/usr` mounted in `/home` before, `/home` typically just has user account directories..

CatFreed: Right now, yes. User accounts haven't been fully implemented yet.  I originally created /home/usr/bin to be the location for user applications because /usr/bin is currently being used for some rootfs applications and will therefore be wiped out every time we update the rootfs. The /home directory is mapped to the new user data partition on the SD card. Everything can be updated, though, if you'd prefer a different layout.

marshall: Makes sense, I'm in favor of the general approach for sure. What do you think about tying this to a user account that we configure for the application, i.e. `/home/$appName/bin`?

CatFreed: We could, but something would need to be done to stick each directory like that in PATH. (Unless we don't want user apps to be accessible through PATH for some reason)

CatFreed:
Update: Finally got sick and tired of not having one, so now there's a progress line instead of a spinner for file transfers.  It looks like this:

  Bytes Sent: 693248/1769379   BPS:8343     ETA 02:08

targets/target-kubos-linux-isis-gcc/target.json
davesims: Subdirectories under `/home` that aren't profiles would be surprising I think. Am I hearing right that `/usr/local/bin` is wiped out, and that's why we can use it?

CatFreed:
@kyleparrott If you could look at the doc update for kubos flash, I'd appreciate it

PR #63, opened 2017-03-06
KUBOS-396 Ping command

kyleparrott:
This implementation is eventually going to get converted into a traditional kubos-linux command project format (when that exists and those builds are supported) rather than this single source file/Makefile configuration.

There's also plans to move the argument parsing to argp when more than one command is implemented  😄

### DIFF
--- a/commands/core.c
+++ b/commands/core.c
@@ -1,0 +1,87 @@
+/*
+ * Copyright (C) 2017 Kubos Corporation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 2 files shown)


commands/core.c
```diff
@@ -0,0 +1,87 @@
...
+
+#include <stdio.h>
+
+#define PING_HASH   6385583923
+#define NUM_ARGS    5   //In the future these values will be inherited from the command
+#define ARG_LEN     20  //and control framework configuration. With this build config that's not simple to do.
+
+int parse_and_run(char * arg);
+
+int execute(int argc, char argv[NUM_ARGS][ARG_LEN])
```
CatFreed: Why do you need to explicitly give the lengths of the arguments?

kyleparrott: There were some unfortunate things previously that required it, but they are gone now. I missed fixing this earlier.

PR #64, opened 2017-03-06
Kubos Arch Diagram

davesims:
### Kubos Architecture Visual Overview

Goal: to give a simple visual overview of Kubos as a whole application. This isn't UML and it isn't totally consistent, but it should provide a simple and intuitive visual that lets us discuss the various subsystems of Kubos and generally how they interact.


CatFreed:
How is pubsub different from telemetry? (might be more of a question for @plauche)

kyleparrott:
This is great, we've definitely needed this!

kyleparrott:
@CatFreed Pubsub is essentially a message bus. It passes publishers and subscribers and can be used independently of telemetry if someone wants. Telemetry is built on top of and uses the pubsub system. It includes specific message formats and other overhead specific for telemetry things.

plauche:
I think the way the pubsub system is represented in this diagram may be an ideal implementation but does not reflect the current state of things. As the code stands now, there is a pubsub api that is used by telemetry to create a telemetry server and allows telemetry clients to interact with it. The diagram appears to indicate a standalone pubsub service that is accessible by the whole system. This pubsub service could be created and operate independently of telemetry for general use, but it currently does that exist in that form.

davesims:
@plauche 👍 makes sense. The goal of this diagram is indeed to represent an _idealized_ Kubos architecture. Would it make more sense to show telemetry as a client of the pub/sub system?

davesims:
Note: the scheduler is shown here as an internal system that queues `job`s that call internal apis. We may end up with a stand-alone scheduler process that calls into C&C, or an internal thread that works in-process, or maybe we'll have call for both. TBD.

CatFreed:
Cool, I like it :smile: The only thing I'm missing is the kind of data being passed between the boxes (ex. what's being passed into the job schedule? Probably schedule requests and then job results), but I don't thing there's an easy way to add that without adding a huge amount of clutter to the diagram.

davesims:
> The only thing I'm missing is the kind of data being passed between the boxes (ex. what's being passed into the job scheduler? Probably schedule requests and then job results), but I don't think there's an easy way to add that without adding a huge amount of clutter to the diagram.

@CatFreed I had the same thought. Not sure yet how to manage that, but it would be great to find a way to call those interfaces out. Even this diagram is too big for the screen and needs to get compressed.

PR #69, opened 2017-03-07
Adding cbor as a csp target dependency

kyleparrott:


targets/target-x86-linux-native/CMake/toolchain.cmake
```diff
@@ -12,3 +12,4 @@ if(NOT (${CMAKE_HOST_SYSTEM_NAME} MATCHES "Linux"))
     message(FATAL_ERROR "This Linux native target will not work on non-Linux platforms (your platform is ${CMAKE_HOST_SYSTEM_NAME}), use `yotta target` to set the target.")
 endif()
 
+set(CMAKE_C_LINK_FLAGS    "${CMAKE_C_LINK_FLAGS} -Wl,-ldl")
```
kyleparrott: @plauche: is this necessary with the addition in the csp CMakeLists.txt? This didn't fix the linking issue and may have unintended consequences down the road.

plauche: We can remove this for now, but the target config is the proper place to change the link flags (once we get libcsp under control).

PR #70, opened 2017-03-09
[KUBOS-340] App init script generation/installation

CatFreed:
Added three new config.json fields:
* system.initAfterFlash - default False. Option for user to specify that their application should be started after being flashed.
* system.initAtBoot - default True. Option for user to specify that their application should be started at boottime
* system.runLevel - default 50 (val 10-99). Sets initialization order value relative to other userland init scripts.

If initAtBoot is set, an init script will be created ("S{runLevel}{appName}") and installed into /home/etc/init.d

Note: In order to work, the kubostech/kubos-linux-build#13 PR needs to be merged.

### DIFF
--- a/targets/target-kubos-linux-isis-gcc/minicom/flash.sh
+++ b/targets/target-kubos-linux-isis-gcc/minicom/flash.sh
@@ -30,10 +30,19 @@ progress() {
 ##########################################################################
 # Minicom doesn't allow any pass-through arguments, so instead we need to 
 # generate a script for it to run.
+# The generate script will:
+#   Navigate to the correct path
+#   Delete any previous versions of the file
+#   Delete any previous init scripts (if flashing application)
+#   Flash the file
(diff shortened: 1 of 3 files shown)


targets/target-kubos-linux-isis-gcc/minicom/flash.sh
```diff
@@ -5,108 +5,247 @@
...
-stty sane
-kill $progress_pid
+##########################################################################
+# Main Script
+##########################################################################
+main() {
+  start=$(date +%s)
+  
+  this_dir=$(cd "`dirname "$0"`"; pwd)
+  file=$1
```
kyleparrott: Does there need to be a check to make sure `file` isn't empty?

kyleparrott: NVM - I see how this script is called..

PR #73, opened 2017-03-13
[KUBOS-350] Linux telemetry service

plauche:
* Executable telemetry service for the linux environment
* Most basic level of functionality, allows for subscribers and publishers, but does not come with any built-in

### DIFF
--- a/linux-telemetry-service/source/main.c
+++ b/linux-telemetry-service/source/main.c
@@ -1,0 +1,35 @@
+#include <stdint.h>
+#include <stdio.h>
+#include <string.h>
+
+#include <ipc/socket.h>
+
+#include <telemetry-linux/server.h>
+#include <telemetry/telemetry.h>
+
+int main(int argc, char ** argv)
+{
+    socket_conn server_conn;
+    kprv_socket_server_setup(&server_conn, TELEMETRY_SOCKET_PORT, TELEMETRY_SUBSCRIBERS_MAX_NUM);
+    bool running = true;
+
+    while (running)
+    {
+        socket_conn conn;
+        while (!kprv_socket_server_accept(&server_conn, &conn))
+        {
+            continue;
+        }
+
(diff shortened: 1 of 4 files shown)


linux-telemetry-service/source/main.c
```diff
@@ -0,0 +1,35 @@
...
+#include <telemetry-linux/server.h>
+#include <telemetry/telemetry.h>
+
+int main(int argc, char ** argv)
+{
+    socket_conn server_conn;
+    kprv_socket_server_setup(&server_conn, TELEMETRY_SOCKET_PORT, TELEMETRY_SUBSCRIBERS_MAX_NUM);
+    bool running = true;
+
+    while (running)
```
CatFreed: How does running get changed? Would setting up a catch for SIGTERM be appropriate?

linux-telemetry-service/source/main.c
```diff
@@ -0,0 +1,35 @@
...
+    {
+        socket_conn conn;
+        while (!kprv_socket_server_accept(&server_conn, &conn))
+        {
+            continue;
```
CatFreed: A sleep might be good, unless there's no risk of the accept taking up too much system time

plauche: The accept is currently a blocking call, so I don't think adding a sleep here will make much difference.

linux-telemetry-service/source/main.c
plauche: I'll add the basic bones (non-functioning) for a more graceful shutdown using the signal catcher, but currently we block on the socket, so we can't really gracefully bring down the process. We should look into reworking the base tcp interface to use a non-blocking connection.

CatFreed: :+1:

linux-telemetry-service/source/main.c
CatFreed: :+1:

PR #74, opened 2017-03-14
KUBOS-376 - Build info command

kyleparrott:
This PR should hopefully be in a state that is good enough to be used as an example for other command libraries. This includes:

- [X] Using argp internally to parse commands and options
- [X] Using a basic `uname` command to get the kernel version.
- [x] Evaluating other command string parsing (Other than string hashes)


kyleparrott:
The tl;dr on the string hashing/comparison with Argp:
 * Argp gives us the ability to parse options and positional arguments when we decide to use them in the future. 
 * Once we get the arguments back from Argp we're still stuck comparing argument strings and mapping those to functions that implement those commands at the end of the day.

commands/source/core.c
```diff
@@ -54,32 +55,27 @@ int help(int argc, char **argv)
...
 }
 
 
-int parse_and_run(char* arg)
+int build_info()
 {
-    unsigned long hash = get_hash(arg);
-    switch (hash)
+    char output [256];
+    char * version_command = "uname --kernel-version";
```
CatFreed: Why'd you choose to call uname like this? It can also be called directly as a function [like this](http://stackoverflow.com/questions/3596310/c-how-to-use-the-function-uname)

commands/source/parser.c
```diff
@@ -0,0 +1,124 @@
...
+            return build_info();
+            break;
+        default:
+            printf("Recevied unknown command: %s\n", command_name);
+            return 1;
+            break;
+    }
+}
+
+
```
CatFreed: int: bunch of extra lines here

commands/source/core.c
kyleparrott: That's really cool and much more simple. I wasn't aware of that.

CatFreed: Glad to help :smile: I ran that example program on the iOBC and the output looks like this:

```
system name = Linux
node name   = (none)
release     = 4.4.23
version     = #12 Thu Mar 2 11:46:35 CST 2017
machine     = armv5tejl
```

PR #71, opened 2017-03-10
KUBOS-398  Moving argument parsing into client

kyleparrott:
The goal of this is to simplify the functionality of the daemon significantly and remove some of the ugly things needed to make the daemon work originally.

This is using the csp fifo interface through named pipes temporarily. This is not ideal, they are only for testing. **They will be removed before review is requested for this PR!** 

 - [x]  Move parsing from daemon into client
 - [x]  Define new Api for sending "Parsed data" into the daemon
 - [x]  Parse the new input data format in the daemon

PR #75, opened 2017-03-14
KUBOS-372 Reboot stub

kyleparrott:
This PR adds a software reboot command to the command and control core library. 

It also fixes a bug in the daemon that was preventing the full length of strings being decoded from Cbor encoded data.


cmd-control-client/source/parser.c
```diff
@@ -93,11 +93,27 @@ bool set_action(char* arg, CNCCommandPacket * command_packet)
...
+    }
+
+    command_packet = state->input;
+
     switch (key)
     {
         case ARGP_KEY_ARG:
+            if (arg == NULL)
+            {
+                return 1;
```
CatFreed: If arg is allowed to be null, you might want to make this a different return code from the other null checks.

commands/source/core.c
```diff
@@ -71,3 +72,15 @@ int build_info()
     printf("Version: %s\n", uname_data.version);
     return 0;
 }
+
+int exec_reboot()
+{
+    int res;
+    printf("Rebooting System\n");
```
CatFreed: According to [the doc](http://man7.org/linux/man-pages/man2/reboot.2.html), "Restarting system" should be printed by the command, so this becomes a duplicate message. Once logging is in place, it could be useful to do something like "Reboot of system triggered by CNC at {date/time}" in a permanent log file.

commands/source/core.c
```diff
@@ -71,3 +72,15 @@ int build_info()
...
+    sync(); //Sync all pending filesystem changes (Logging)
+    if (res = reboot(RB_AUTOBOOT) != 0)
+    {
+        printf("There was an error rebooting the system.\n");
```
CatFreed: Print the return code here

cmd-control-client/source/parser.c
kyleparrott: I've addressed this in #71

PR #78, opened 2017-03-22
KUBOS-414 Moving daemon specific defines out of the command-and-control into daemon. Misc fixes for running on the iOBC

kyleparrott:
This:
* Removes a couple of hard coded dev paths
* Moves some daemon specific defines out of the command-and-control module, and into the daemon
* Adds Config.jsons specific for getting the command and control framework running on the iOBC.

### DIFF
--- a/cmd-control-daemon/cmd-control-daemon/daemon.h
+++ b/cmd-control-daemon/cmd-control-daemon/daemon.h
@@ -47,6 +47,16 @@ bool cnc_daemon_start_encode_response(int message_type, CNCWrapper * wrapper);
 bool cnc_daemon_send_result(CNCWrapper * wrapper);
 
 
+#define LIB_FORMAT_STR "/lib%s.so"
+
+#ifndef YOTTA_CFG_CNC_REGISTRY_DIR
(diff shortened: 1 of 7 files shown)


cmd-control-daemon/cmd-control-daemon/daemon.h
```diff
@@ -62,6 +72,6 @@ bool cnc_daemon_send_result(CNCWrapper * wrapper);
 #ifdef YOTTA_CFG_CNC_DAEMON_LOG_PATH
 #define DAEMON_LOG_PATH YOTTA_CFG_CNC_DAEMON_LOG_PATH
 #else
-#define DAEMON_LOG_PATH "/home/vagrant/daemon.log" //Only a dev path for now
+#define DAEMON_LOG_PATH "/var/log/daemon.log"
```
CatFreed: Everything that's not under /home or /upgrade will be wiped during a firmware upgrade/rollback. I think it'd be better to put logging somewhere under the /home directory. (/home/var/log would probably work. You'd have to make sure you create the directory, though, because it doesn't currently exist)

PR #83, opened 2017-03-28
Pushing only the new tag rather than all tags in the CD deploy script

kyleparrott:
This should only push the new CD tag and prevent any janky tag conflicts like [build 171](https://circleci.com/gh/kubostech/kubos/171)


kyleparrott:
@davesims: Friendly request for 👀

PR #85, opened 2017-03-30
KUBOS-435 Kubos-SDK Doc updates

kyleparrott:
This addresses some formatting, spelling, and new options that haven't been documented in the sdk docs.

### DIFF
--- a/docs/sdk-installing.md
+++ b/docs/sdk-installing.md
@@ -1,6 +1,6 @@
 # Installing Kubos SDK
 
-## What is the Kubos SDK?
+## What is the Kubos SDK? {#what-is-kubos-sdk}
 
 The Kubos SDK is a term used to describe the all of the components used that make up a KubOS operating system and the tools used to build this project.
 The Kubos SDK components are made up of:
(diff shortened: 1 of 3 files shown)


docs/sdk-installing.md
```diff
@@ -1,6 +1,6 @@
 # Installing Kubos SDK
```
CatFreed: In order to get your headers with "{#_blah_}" to display, you need to add a second top-level header with it's own "{#-_blah_}". So:
```
# Installing Kubos SDK
# Installing Kubos SDK {#installing-kubos-sdk}
```

docs/sdk-installing.md
```diff
@@ -1,6 +1,6 @@
...
 
-## What is the Kubos SDK?
+## What is the Kubos SDK? {#what-is-kubos-sdk}
```
CatFreed: Why did you add these?

docs/sdk-reference.md
```diff
@@ -1,44 +1,28 @@
...
-
-[use](#kubos-use)                 Set a new version of the Kubos modules to build your projects against.
+| Command                          | Function                                              |
+| ------------------------------   | ----------------------------------------------        |
+|[build](#kubos-build)             | Build the current module.                             |
+|[clean](#kubos-clean)             | Remove files created by kubos and the build.          |
+|[config](#kubos-config)           | Display the target configuration info.                |
+|[debug](#kubos-debug)             | Attach a debugger to the current target.  Requires target support.       |
+|[flash](#kubos-flash)             | Launch the compiled program (available for executable modules only). Requires target support for cross-compiling targets. |
+|[flash](#kubos-flash-linux)       | (KubOS Linux) Load files onto target.                 |
```
CatFreed: Nit: change to "KubOS Linux targets" to match your other change

docs/sdk-reference.md
```diff
@@ -216,41 +200,49 @@ Note: This requires target support.
 Pull and update all of the current Kubos modules. By default if no `<version number>`
 
 ### Arguments
```
CatFreed: Is there a difference between "Arguments" and "Options"? If not, change things so we're only using one of them.

docs/sdk-upgrading.md
```diff
@@ -2,7 +2,7 @@
 
 New updates will be announced on the [Kubos website](http://docs.kubos.co/). There will be instructions depending on the specific release details of all the steps needed to upgrade for the newest release.
 
-## Upgrading Kubos CLI
+## Upgrading Kubos CLI {#upgrade-kubos-cli}
```
CatFreed: Same comments/updates as sdk-installing.md

docs/sdk-installing.md
kyleparrott: Because they're useful for support things. 

There's a lot less that can go wrong when you send someone a link to a specific section of a page rather than just to the top of the page and then hope that they scroll down to the right section. 

Since this is the most complex part of setting up a kubos dev environment I don't see the harm in adding them.

docs/sdk-upgrading.md
kyleparrott: As above - they can be useful but if you feel strongly them i'll remove them.

CatFreed: Ah, I meant that you need the extra top-level header line in order for these to display properly

docs/sdk-installing.md
CatFreed: Cool, works for me

PR #86, opened 2017-03-30
[KUBOS-445] Userland supervisor interface

plauche:
- [x] Verify that low level SPI commands work (this will be confirmed once we establish supervisor communications)
- [x] Fully verify low level comms with the CRC
- [x]  Replicate iOBC supervisor functionality to work in userland linux
  - Currently using ioctl calls directly in the supervisor code
- [ ] ~(Stretch) modify kubos-hal-linux to perform actual linux spi abstraction~ (Saving this for later)
- [x] Document iobc hal

PR #88, opened 2017-04-05
[KUBOS-425] Supervisor telemetry integration

plauche:
Pending integration with the iOBC supervisor chip #86, we would like to start exposing metrics from the supervisor into the telemetry system.

- [ ] Create publisher for supervisor's uptime metric
- [ ] Create publisher for supervisor's reset count metric 
- [ ] Integrate publishers into on-board telemetry service


linux-telemetry-service/source/supervisor_publisher.c
```diff
@@ -0,0 +1,63 @@
...
+
+CSP_DEFINE_TASK(supervisor_publisher)
+{
+    while (1)
+    {
+        supervisor_housekeeping_t info;
+        if (supervisor_get_housekeeping(&info))
+        {
+            telemetry_publish((telemetry_packet) {
+                .data = info.fields.iobcUptime,
```
CatFreed: Consider converting 'iobcUptime' and the other fields to lower case with '_' seperators to match our coding standards. I know these names originally come from ISIS, though.

linux-telemetry-service/config.json
```diff
@@ -6,6 +6,12 @@
         },
         "subscribers": {
             "max_num": 10 
+        },
+        "publishers": {
+            "iobc_uptime": 100,
+            "supervisor_uptime" : 101,
+            "iobc_reset_count" : 102,
+            "iobc_serial": 103
```
CatFreed: Why would someone want to subscribe to watch the serial number? In theory it should always be a static value.

plauche: Haha yea true, I'll remove that.

PR #92, opened 2017-04-07
[KUBOS-420] KubOS Linux Recovery Architecture doc

CatFreed:
Created a doc to cover the various steps that are taken if something bad happens to the KubOS Linux kernel and the system fails to boot succesfully too many times.

Misc: Also fixed uclibc->glibc change that was missing


docs/kubos-linux-recovery.md
```diff
@@ -0,0 +1,92 @@
+# KubOS Linux Recovery Architecture
```
plauche: Double title here

docs/kubos-linux-recovery.md
```diff
@@ -0,0 +1,92 @@
...
+
+If for some reason your KubOS Linux system boots after an upgrade but has introduced some non-critical issue (like an incompatibility with a user application), you can manually rollback to a previously installed version. Previous packages are not deleted once they have been loaded. As a result, you can simply specify which package you would like to boot into and then restart your system.
+
+From the KubOS Linux shell:
+
+    $ fw_printenv kubos_updatefile kpack-{desired version}.itb
+    $ reboot
+
+
+## Boot into the Alternate OS
```
CatFreed: No, and I'm not sure if we should, what it should be, and/or where we should put it. It's super non-portable. For the iOBC, it's 1) Create an iOBC RT application, 2) flash to any sector at offset 0x00090000 or greater, 3) update the default iobc altbootcmd with the address and length of the application.

I'll add a small generic summary.

docs/kubos-linux-recovery.md
CatFreed: It's actually a [requirement of Doxygen](https://www.stack.nl/~dimitri/doxygen/manual/markdown.html#md_page_header). The first level 1 header is used as the page title and the second is used as the actual header. For this particular document it doesn't matter, because there aren't any of the header labels ("## Header Whatever {#header-whatever}". However, I've found that it's useful to go ahead and include the second header anyways, in case header labels are added in the future, so that we don't accidentally run into formatting problems.

docs/main.md
```diff
@@ -44,8 +44,9 @@ If for some reason Slack won't work for you, feel free to email us at info@kubos
 
  - [User Applications on iOBC](docs/user-app-on-iobc.md)
  
- - [Upgrade KubOS Linux](docs/kubos-linux-upgrade.md)
-
+ - [KubOS Linux Upgrades](docs/kubos-linux-upgrade.md)
+  
+ - [KubOS Linux Recovery](docs/kubos-linux-recovery.md)v
```
kyleparrott: Nit: "v" at the end of the line.

PR #94, opened 2017-04-11
Linking doxygen projects

plauche:
Doxygen allows for the generation of "tag files" which will inform projects about the structure and data of other projects. This allows doxygen to create automatic and manual links across projects.

* New script, `tools/gendocs.py`, which facilitates generating the tag and html files
* Updated top level `main.md` to use file references instead of absolute paths
* Created reference anchors in `main.md` files for referenced projects


tools/gendocs.py
```diff
@@ -0,0 +1,88 @@
...
+echo "HTML_OUTPUT=.";
+echo "EXCLUDE=source test";
+echo "EXTERNAL_GROUPS=NO";
+echo "EXTERNAL_PAGES=NO";) | doxygen -"""
+
+DOC_TAG_DIR = """
+{0}/output.tag={1}
+"""
+
+DOCS_DIRS = [
```
kyleparrott: Can we auto generate this list so we don't have to edit this each time we add a new module with a docs directory, or do you think that's overkill?

tools/gendocs.py
```diff
@@ -0,0 +1,88 @@
...
+".",
+"kubos-core", 
+"libcsp", 
+"freertos/os", 
+"hal/isis-iobc-hal",
```
CatFreed: Note: This folder doesn't exist in `master` yet and will cause this script to fail. I know that it's sitting in one of your other PRs, just wanted to point out the slight dependency

tools/gendocs.py
plauche: Hmmm we potentially could. We just want to make sure that we aren't generating docs for every folder, as some don't even have a doxygen setup (like all of the STM32CubeF4 folders), and some do but maybe we don't want to generate their docs (tinycbor).

hal/kubos-hal-msp430f5529/docs/main.md
```diff
@@ -1,4 +1,5 @@
 # Kubos-HAL-MSP430F5529
+## {#kubos-hal-msp430f5529-main}
```
plauche: For some weird doxygen reason that is what it took for these labels to link correctly through the tag files. They don't actually show up in the rendered html though.

tools/gendocs.py
plauche: Ahhh shoot, yup I will remove it.

PR #95, opened 2017-04-11
[KUBOS-239] Adding info about users, directories, and user permissions

CatFreed:
Original goal was to document the location of the device_table.txt file, which contains the default user permissions in regards to the systems libraries (/tmp, /root, /etc, etc), as well as the basic process for adding more users.

Realized that I hadn't actually documented the userspace directories either, so I threw them in there, too.


docs/kubos-linux-overview.md
```diff
@@ -120,3 +120,45 @@ Final Distribution: \*.dtb file.
...
+
+There are a few key directories residing within the KubOS Linux user space
+
+#### /usr/sbin
+
+All built-in Kubos services will reside in the /usr/sbin directory. This covers things like the telemetry and command and control services.
+
+#### /home
+
+All user-created files should reside under the /home directory. This directory maps to a separate partition from the root file system. As a result, all file here will remain unchanged if the system goes through a kernel upgrade or downgrade.
```
kyleparrott: nit: file -> files

docs/kubos-linux-overview.md
```diff
@@ -120,3 +120,45 @@ Final Distribution: \*.dtb file.
...
+
+***
+**Any files not residing under the /home directory will be destroyed during an upgrade/downgrade**
+***
+
+##### /home/usr/bin
+
+All user-created applications will be loaded into this folder during the `kubos flash` process. The directory is included in the system's PATH, so applications can then be called directly from anywhere, without needing to know the full file name.
```
kyleparrott: perhaps "the full file path" is more accurate than "the full file name"

docs/kubos-linux-overview.md
```diff
@@ -120,3 +120,45 @@ Final Distribution: \*.dtb file.
...
+
+### File System
+
+There are a few key directories residing within the KubOS Linux user space
+
+#### /usr/sbin
+
+All built-in Kubos services will reside in the /usr/sbin directory. This covers things like the telemetry and command and control services.
+
+#### /home
```
plauche: If we are creating a specific user account, we may want to point out that user's home folder as the preferred place for user-created files.

docs/kubos-linux-overview.md
```diff
@@ -120,3 +120,45 @@ Final Distribution: \*.dtb file.
...
+
+#### /home/usr/bin
+
+All user-created applications will be loaded into this folder during the `kubos flash` process. The directory is included in the system's PATH, so applications can then be called directly from anywhere, without needing to know the full file path.
+
+#### /home/usr/local/bin
+
+All user-created non-application files will be loaded into this folder during the `kubos flash` process. There is currently not a way to set a destination folder for the `kubos flash` command, so if a different endpoint directory is desired, the files will need to be manually moved.
+
+#### /home/etc/init.d
```
plauche: Should this be `/home/usr/etc/init.d`?

docs/user-app-on-iobc.md
```diff
@@ -140,6 +140,24 @@ Assuming you've successfully built a Kubos SDK project for the ISIS-OBC board, w
     /
     Transfer Successful
     Execution time: 21 seconds
+    
+The application binary will be loaded into the /home/usr/bin directory on the target board.
+
+If the 'system.initAtBoot' option has been turned on, then a standard initialization script will be generated and flashed into the /home/etc/init.d directory automatically during the application flashing process.
```
plauche: Where are the `system.initAtBoot` and `system.initAfterFlash` options set/stored?

docs/kubos-linux-overview.md
CatFreed: No, it's currently /home/etc/init.d. The intention with this (and with /home/usr/bin) was to take whatever the root system directory is (so /etc/init.d and /usr/bin) and simply place it under the /home directory.

docs/user-app-on-iobc.md
CatFreed: They're config.json options (documented in the "Setting Initialization Configuration" section of this doc)

PR #97, opened 2017-04-12
Enabling search on main kubos docs page

plauche:
We had previously disabled the search box on the main docs page, this turns it back on. The search box on the top level docs page (and on the sub project pages) *should* search across all projects due to the project linking.

* Does _not_ search through our markdown docs


CatFreed:
Is there not an easy way to search the markdown docs?

PR #112, opened 2017-04-18
[KUBOS-469] Standards Doc Updates

CatFreed:
Updating our standards doc to contain actual, fleshed out standards.

Also updated the .clang-format file to match these new standards.

Questions:
- C
  - Implicit conversion to bool in conditions? ("if (ret)" vs "if (ret != NULL)")
  - Each variable in own declaration?
  - Returning: only allow return codes, or allow return values as well?
  - Comments: only allow C style (/* */), or allow C++ style, too (//)

@kyleparrott, @plauche, @davesims, or really, anyone who actually knows and uses Python. I just put a link to the pep8 standard. I don't if we want to go into any more detail than that...

### DIFF
--- a/docs/kubos-standards.md
+++ b/docs/kubos-standards.md
@@ -108,22 +108,17 @@ This section covers the styling and standards for the various languages and tool
 
 ## C {#c-standards}
 
-[ClangFormat](https://clang.llvm.org/docs/ClangFormat.html) is a series of tools that can be used to automatically correct any C coding inconsistencies. You can find an example which we've used in the '.clang-format' file in the [Kubos repo](https://github.com/kubostech/kubos/blob/master/.clang-format)
+[ClangFormat](https://clang.llvm.org/docs/ClangFormat.html) is a series of tools that can be used to automatically correct any C coding inconsistencies. 
+A stand-alone tool is available, which can then be used with a variety of IDEs. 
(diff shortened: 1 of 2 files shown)


docs/kubos-standards.md
```diff
@@ -1,55 +1,343 @@
...
+[TOC]
 
-##Product Names
+# Kubos Coding and Documentation Standards {#kubos-standards}
 
-The general naming scheme is "Kub[OS|os] {component}".  Note that there is a space separating the two words.
+This is a doc to maintain the current naming and coding standards when working with the Kubos products.
 
-If the component is an OS, then use the capitalized "OS".  If not, then use "os".
+# Them's Fightin' Words {#fightin-words}
```
kyleparrott: :cowboy_hat_face:

docs/kubos-standards.md
```diff
@@ -1,55 +1,343 @@
...
+
+Use name prefixes for identifiers declared in different modules.
+For example, 'csp\_buffer\_free' indicates that the function belongs to the CSP directory.
+
+## Python {#python-standards}
+
+[Pylint](https://pylint.readthedocs.io/en/latest/) is a great tool which can be used to check the style and validity of your python files.
+It has support for a variety of [editors and IDEs](https://pylint.readthedocs.io/en/latest/user_guide/ide-integration.html).
+
+Refer to [Python's Python Style Guide](https://www.python.org/dev/peps/pep-0008/) for our preferred Python styling.
```
kyleparrott: This is fine by me as far as documenting the python styling.

PR #115, opened 2017-04-24
[KUBOS-476] Converting user docs to reStructuredText

CatFreed:
Converting user docs from markdown to reStructuredText since it's a much more standardized documentation style.

Notes: 
1. ":doc:" and ":ref:" tags don't appear to be rendered correctly in github. They do create the correct file links when externally generated, though.
2. This PR is **NOT** going into master. It's currently being applied against @plauche's branch, because there is still some work to be done for the API doc generation side of things.


docs/index.rst
```diff
@@ -1,29 +1,82 @@
...
+Looking to build an application on Kubos? Check out our :doc:`quick start
+guide <first-project>`.
+
+Having issues? Want a new feature? `Come talk to
+us! <https://slack.kubos.co/>`__
+
+If for some reason Slack won't work for you, feel free to email us at
+info@kubos.co.
+
+.. toctree::
```
kyleparrott: These blocks render kind of janky in the rich diff. I'm assuming that that's a github RST rendering issue?

CatFreed: Appears that way. The ':hidden:' tag, as the name would imply, is supposed to hide the table of contents on the page. It works fine when you actually gen the docs. The github-generated format isn't particularly unsightly though, so I'm inclined to leave it as-is for now.

PR #117, opened 2017-04-27
[KUBOS-478] Migrating example projects to new examples folder

CatFreed:
Since all of the other kubos subcomponents (core, hal, csp, etc) have been moved to live under the main kubos repo, the example projects should live here, too.

Projects have been migrated using `git subtree` to attempt to preserve their histories. If there's a better approach, let me know and i'll change things.


kyleparrott:
Should the `kubos-shell-example` be added yet since #34 is not merged yet, and won't be usable yet?

CatFreed:
The example already exists in its own repo, so it's not like it's new content. Also, doing it now with the other examples means no one has to remember what was done to move it later.

PR #120, opened 2017-04-28
[KUBOS-479] Document Examples

CatFreed:
Created a new doc to go over all the example applications that we have.

The links to the various examples won't work until #117 has been merged.

Note: The final example (kubos-shell-example) is currently in a 'todo' section because the Slash code (#34 from @marshall ) has not been merged yet, so the example won't work out-of-the-box. Github displays the 'todo' directive, but the final doc gen with sphinx will keep it hidden.


TheHollidayInn:
Sweet! I think it might be useful to define some of the acronyms in the descriptions. And maybe link to more information?

CatFreed:
@TheHollidayInn I can definitely do that :smile: Can you give me an example of the kind of information you'd like to see?

TheHollidayInn:
Sure! For example: https://github.com/kubostech/kubos/pull/120/files#diff-a5f2dcd2f41415a34902f8ebddc864d9R103 - I'm unsure of what CSP stands for or how to learn about it. A quick Google search, and I see at least two kinds of CSP servers. 

Granted, I do have a strong understand of Embedded systems - so it might be a common acronym.

CatFreed:
I added reference links to each of the high-level items in the 'components used' tables. This should give a basic description of them, as well as a link to a relevant doc.

Small catch: we're in the process of reorganizing our docs. The link locations are where the docs likely will be, not where they currently are. Also, the SDK config doc doesn't exist yet (but is in the backlog to be created soon.)

TheHollidayInn:
Awesome! I think that helps a lot. Thank you :D

docs/sdk-examples.rst
```diff
@@ -0,0 +1,283 @@
...
+2. If the `BME280 sensor <https://cdn-shop.adafruit.com/datasheets/1899_HTU21D.pdf>`__ has been defined in the project's config.json file, the appropriate initialization calls will be made and then the application will enter a loop.
+   
+   In each iteration of the loop, the application will fetch and print the current temperature and humidity data from the sensor to the default UART port.
+   
+Kubos SD Example
+----------------
+
+`kubos-sd-example <http://github.com/kubostech/kubos/examples/kubos-sd-example>`__
+
+**Compatible Targets: MSP430F5529, STM32F407 Discovery, STM32F405**
```
kyleparrott: Do we want users actually building with the `STM32f405` target? I thought it was only used as a parent target for the `na-satbus-3c0` and the `pyboard-gcc` targets.. which both implement their own hardware specific flashing functions which are different than the flash routine implemented by the `STM32f405` target.

CatFreed: @plauche Can you weigh in on this. I'm not really familiar with the history of all the boards...
I added the STM32F405 here because its target.json file defines an available SPI bus, which is what's needed to then use the SD card

plauche: Shoot I missed that. Yea the stm32f405 target is a high/mid level target, it does not actually correspond to one specific board.  The pyboard and na-satbus-3c0 are the two board specific targets we have that are derived from the stm32f405.

CatFreed: :+1:

PR #125, opened 2017-05-04
[KUBOS-504] Docs cleanup kubos-hal-stm32f4

plauche:
* A bit of cleanup on the gendocs.py tool
* Adds options to include structures on api pages
* Standardizing capitalization
* Added missing api level docs


hal/kubos-hal-stm32f4/source/uart.c
```diff
@@ -320,9 +322,9 @@ void kprv_uart_enable_tx_int(KUARTNum uart)
 }
 
 /**
- * Write a character directly to the uart interface
- * @param uart uart bus
- * @param c character to write
+ * Write a character directly to the UART interface
```
CatFreed: Might be nice to mention that 'directly' means that it bypasses the tx queue

PR #129, opened 2017-05-11
KUBOS-520 Adding Kubos CI Dockerfile

kyleparrott:
Mirrors the Vagrant environment only for standard kubos projects (no linux dependencies).

This is a slightly stripped down (the linux dependencies have been removed) image that has been running in our CI builds for the kubos and kubos-linux-build repos.

### DIFF
--- a/tools/dist/programs.json
+++ b/tools/dist/programs.json
@@ -1,0 +1,8 @@
+{
+    "programs": [
+        "msp430-gcc",
+        "arm-none-eabi-gcc",
+        "arm-none-eabi-gdb",
+        "arm-linux-gcc"
+    ]
+}
--- a/tools/dist/Dockerfile
+++ b/tools/dist/Dockerfile
@@ -35,7 +35,8 @@ RUN rm ./iobc_toolchain.tar.gz
 
 RUN pip install pysocks
 RUN pip install mock
-RUN pip install kubos-cli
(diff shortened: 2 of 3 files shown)


kyleparrott:
They are the same versions (for the 2 arm compilers and the msp compiler). I've added a pretty basic script for checking the versions of installed utilities between the docker container and a given Vagrant environment to make this easier in the future.

tools/dist/programs.json
```diff
@@ -0,0 +1,8 @@
+{
+    "programs": [
+        "msp430-gcc",
+        "arm-none-eabi-gcc",
+        "arm-none-eabi-gdb",
+        "arm-linux-gcc"
```
CatFreed: What are these programs for? It appears to only be a small subset of our current targets

kyleparrott: Aside from the `arm-none-eabi-gdb` program, which was left in by mistake, they're the main 3 gcc instances from the toolchains we use. They're a response/effect of @plauche's comment:

> Have we verified that the cross compilers for arm and msp430 are the same versions here and in the vagrant environment? 

They feed into the script to automate checking versions of programs installed  in the vagrant and docker environments to ensure that the docker environment is as close of a mirror of the vagrant environment as possible. Since the docker instances are only ever used to compile programs that's why only the "gcc's" are checked.

PR #130, opened 2017-05-16
KUBOS-522 Installation script for host

kyleparrott:
This script installs:
* virtualbox
* virtualbox extension pack
* vagrant
* vagrant vb-guest plugin
* downloads the latest kubos-dev vagrant box

This is intended to work through a command like `curl <url of this script> | /bin/bash`

It has been tested on the latest versions of Centos, Fedora and Ubuntu, Debian and Macos. 

I am open to input, especially with how to handle running some commands as root and others not.


tools/get-kubos/install.sh
```diff
@@ -0,0 +1,196 @@
...
+virtualbox
+)
+
+install_list=()
+
+set_platform() {
+    platform=$(uname)
+    if [[ $platform == "Darwin" ]]; then
+        os="macos"
+        pm="brew"
```
davesims: Can we expand this to `package_manager`?

davesims: Can we expand `pm` to `package_manager`?

PR #131, opened 2017-05-17
KUBOS-523 Updating docs for Github now hosting CLI wheel releases

kyleparrott:


docs/sdk-reference.rst
```diff
@@ -310,9 +310,12 @@ Options
...
    then kubos will try to checkout the provided version number after
-   pulling the latest updates
--  ``--latest``, ``-l`` Checkout the latest release during the update
-   process.
+   pulling the latest updates.
+- ``--all``, ``-a`` Update the Kubos source modules and the Kubos CLI python module.
+- ``--cli``, ``-c`` Download and update the Kubos CLI python module only.
+- ``--latest``, ``-l`` Checkout the latest release during the update process.
+- ``--source``, ``-s`` Only update the source modules. This is the default if no other options are specified.
+- ``--tab-completion``, ``-t`` Update the tab completion definitions. This option is only necessary after updating the CLI.
```
CatFreed: I'm kind of entertained at the length of this command. I'm gonna want to tab complete the tab-completion option...

docs/sdk-upgrading.rst
```diff
@@ -20,11 +20,18 @@ First SSH into your kubos-dev box:
        $ vagrant up
        $ vagrant ssh
 
-The Kubos CLI can be upgraded using this pip command:
+The Kubos CLI can be upgraded using this kubos command:
```
CatFreed: This sentence uses a different grammar structure than the previous line. Change this one to something like "Then, upgrade the Kubos CLI"

docs/sdk-upgrading.rst
```diff
@@ -20,11 +20,18 @@ First SSH into your kubos-dev box:
...
 
-The Kubos CLI can be upgraded using this pip command:
+The Kubos CLI can be upgraded using this kubos command:
 
 ::
 
-        $ sudo pip install --upgrade kubos-cli
+        $ kubos update --cli
+
+New releases of the Kubos CLI may add or remove specific commands or command options.
```
CatFreed: Why isn't the tab completion table updated when you update the CLI? Seems like something that could be bundled.

docs/sdk-upgrading.rst
```diff
@@ -20,11 +20,18 @@ First SSH into your kubos-dev box:
...
+To update the tab completion options in your Kubos Vagrant box run the following command:
```
CatFreed: I don't like this sentence, but I can't figure out why or what it should be instead. Maybe something like "To keep tab completion synchronized with the Kubos CLI, run the following command:"? Not sure...

CatFreed: Alternatively, maybe we just need a blurb about what tab completion is/is for

docs/sdk-upgrading.rst
kyleparrott: It will be integrated into the installation at some point - but not right now.

That is a near future goal. It requires a little bit of some funky installation setup that is going to require a fair amount of testing, which just hasn't been done yet. Some of the python installation configurations are pretty fragile and delicate. I want to be very confident in the module's installation configuration when that feature gets shipped.

Issue #121, opened 2017-05-01
[KUBOS-486] Quickstart Tutorial Video


CatFreed:
I created a quickstart video, located in [our engineering folder](https://drive.google.com/file/d/0B5Ykyv8ax7fUeGVaXzIwRWJtRUU/view?usp=sharing).
This issue has been created to house the video review process. Once this review is complete, the video will be added to the docs...somewhere...

@kubostech/devs

CatFreed:
Inserted video into the main page and created a PR: #132

PR #132, opened 2017-05-22
[KUBOS-486] Adding quick start tutorial to main page

CatFreed:
Created new docs/_static folder to house the tutorial video. Might want to pursue hosting videos elsewhere in the future.

Created in conjunction with issue #121 

Misc:
  - Fixing standards doc headers to have correct levels
  - Fixing random typo in upgrade doc


TheHollidayInn:
What about hosting them on an S3 server? It is pretty cheap.

kyleparrott:
I'd propose hosting them either as (un)published videos on youtube and embedding them in our docs that way. Their embedded player is pretty nice and we don't have to mess with all the other
 hosting issues around it 😄

CatFreed:
The video is now hosted on YouTube. It might get migrated to AWS in the future, since there's some other things we'd like to host, but this will work for now.

PR #133, opened 2017-05-30
[KUBOS-518] Updating linux docs

CatFreed:
- Updating the iOBC build instructions to mention need for `sudo`.
- Adding instructions for changing toolchain destination.
- Adding warning about inability to build KubOS Linux using shared folders
- Updating altbootcmd/alternate OS explanation


CatFreed:
@kyleparrott We definitely should. @plauche Is the script/process to generate the image done? Or done enough to write up how it works?

kyleparrott:
I think it should be pretty much good to go. The use case I was talking about where user's not actually building the image themselves. Their process would be simple:

1) Download [Etcher](https://etcher.io/) or whatever flashing tool we want to officially recommend.
2) Download the latest image
3) Flash the sd card
4) 💰 💰 💰 

The only thing we need at this point is the link to where we're going to host the image at. I'm fine with leaving the specific link as a TODO until we figure out the hosting details.

---

For those who just really want to build the kernel themselves, the image generation script can get baked into the buildroot defconfig `BR2_ROOTFS_POST_IMAGE_SCRIPT` option so it  just runs after the build is complete.

CatFreed:
@kyleparrott :+1: 
So, future to-do's:
- Add image download instructions (once the image download location has been decided)
- Update `BR2_ROOTFS_POST_IMAGE_SCRIPT` to point to our new image creation script

docs/kubos-linux-on-iobc.rst
```diff
@@ -466,7 +659,7 @@ You should see the console boot into Linux like this:
    :alt: Linux Console
 
    Linux Console
-
+
```
kyleparrott: Nit: Rogue space

PR #141, opened 2017-06-09
Migrating U-Boot Env: Doc updates + misc

CatFreed:
Tweaking iOBC NOR flash addresses to match changes in kubostech/uboot#12 

Also:
- Fixing default system.destDir to use our /home/system file structure
- Changing default system.initAtBoot to be false


CatFreed:
[Migrate U-Boot envars to SD card](https://trello.com/c/wxszCkfx/36-migrate-u-boot-envars-to-sd-card)

PR #142, opened 2017-06-12
Integration test runner

kyleparrott:
This implements the "Integration Test Runner Script" and basic examples for the:

* Hello world script test. - Verifies non-project file uploads
* Hello world compilation test - Verifies project build and upload
* c2 ping test - Verifies that c2 is setup and active

There is a readme in this PR that describes all of the Test Runners configuration options in detail.


kyleparrott:
[Master QA Script](https://trello.com/c/fn6oi8gz/44-master-qa-script)

kyleparrott:
[Hello World Test Project](https://trello.com/c/di3ijEEm/43-hello-world-test-project)

test/integration/linux/config.py
```diff
@@ -0,0 +1,70 @@
+import os
+import json
+import sys
+
+class KubosTestConfig(object):
+    tests = []
+    class Device(object):
+        def __init__(self):
+            self.buadrate = 115200
```
davesims: s/buadrate/baudrate/

test/integration/linux/config.py
```diff
@@ -0,0 +1,70 @@
...
+            self.prompt = None
+            self.target = None
+            self.timeout = 10
+
+
+    class Login(object):
+        def __init__(self):
+            self.username = 'root'
+            self.password = 'Kubos123'
```
davesims: Do these need to go into `ENV`s?

davesims: ...or in the config read in below?

test/integration/linux/hello-world/README.md
```diff
@@ -0,0 +1,6 @@
+# KubOS Linux Hello World
```
davesims: Is this still needed?

test/integration/linux/config.py
davesims: ...or more likely, extracted into 

```python
DEFAULT_USERNAME = 'root'
DEFAULT_PASSWORD = 'Kubos123'

[...]

self.username = DEFAULT_USERNAME
self.password = DEFAULT_PASSWORD
```

test/integration/linux/hello-world/README.md
kyleparrott: Not especially right now. But future documentation could be added about what this project is used for and the test cases it is used to verify.

I don't have strong feelings either way.

test/integration/linux/test_runner.py
```diff
@@ -0,0 +1,269 @@
...
+        $ kubos flash <path-to-file>
+
+        Which must be executed from a linux project directory.
+        '''
+
+
+        proj_name = 'dummy-project'
+        temp_dir = tempfile.mkdtemp()
+        os.chdir(temp_dir)
+        self.run_cmd('kubos', 'init', '-l', proj_name)
```
CatFreed: Turns out you have to have a "build/{target}" folder within your project in order for flashing to work (the actual build process doesn't have to succeed though)

test/integration/linux/readme.md
```diff
@@ -0,0 +1,73 @@
...
+
+### Tests
+
+This section is an array of test descriptions. The tests are run in the order they are listes in the array.
+
+* name - The name of the test. For user recognition only. This has no effect on the test.
+* pre_test - A shell command that will be executed before cloning, building, or running any tests. This attribute is not required
+* build_source - A description of where the test source is located. This can be a local directory or a git repo url. This section is not required. If not provided, no binaries will be built.
+* flash_source - The path to what should be flashed to the target. If you have a build source specified this attribute is not necessary. It is intended to be used to upload non-built resources (like shell scripts)
+* test_command - The acutal command that will be executed on the board to start the test on the target device.
```
plauche: nit: spelling `actual`

test/integration/linux/test_runner.py
```diff
@@ -0,0 +1,269 @@
...
+
+
+    def create_dummy_project(self):
+        '''
+        This function creates a dummy linux project specifically for flashing
+        non-project files to the target devices. Currently this is done by:
+
+        $ kubos flash <path-to-file>
+
+        Which must be executed from a linux project directory.
```
plauche: Do we have a task to enable flashing files outside of a project? Is that something we want to consider?

test/integration/linux/test_runner.py
```diff
@@ -0,0 +1,269 @@
...
+        flash_args = ['kubos', '-t', self.config.device.target, 'flash']
+
+        if not os.path.isabs(proj):
+            proj = os.path.normpath(os.path.join(start_dir, proj))
+            self.logger.info('Interperating provided path as relative. Using %s' % proj)
+
+        if os.path.isfile(proj):
+            # flash the project as a standalone file
+            flash_args.append(proj)
+            proj_dir = self.create_dummy_project()
```
plauche: Does the dummy project you are flashing from need to be set to the correct target prior to flashing?

test/integration/linux/test_runner.py
kyleparrott: Yepp, but post 1.0 I believe.

test/integration/linux/generic_test_config.json
```diff
@@ -0,0 +1,32 @@
...
+        "timeout": 2
+    },
+    "login": {
+        "username": "root",
+        "password": "Kubos123"
+    },
+    "tests": [
+        {
+            "name": "Hello World Script Upload Test",
+            "test_command" : "/root/hello_world.sh",
```
davesims: @kyleparrott @CatFreed where is the default destination set?

test/integration/linux/generic_test_config.json
```diff
@@ -0,0 +1,32 @@
...
+            "flash_source": "./hello-world/script/hello_world.sh",
+            "expected_test_output": "Hello World!"
+        },
+        {
+            "name": "Hello world Compilation Test",
+            "test_command" : "/usr/sbin/hello_world",
+            "build_source": "/Users/kyleparrott/code/kubos/kubos/test/integration/linux/hello-world",
```
davesims: Should be from a config or `ENV` more than likely?

test/integration/linux/test_runner.py
```diff
@@ -0,0 +1,269 @@
...
+        self.logger.info("Running test: %s" % test.name)
+
+        if test.pre_test:
+            self.logger.info("Test: %s Running pre-test command: %s\n" % (test.name, test.pre_test))
+            self.run_cmd(test.pre_test)
+
+        #build the project
+        if test.build_source:
+            resource_type = self.get_resource_type(test.build_source)
+            logging.info("Test %s has type: %s" % (test, resource_type))
```
davesims: yeah looks like the default serialization for that object isn't friendly.

test/integration/linux/test_runner.py
```diff
@@ -0,0 +1,269 @@
+#!/usr/bin/env python
+
+import argparse
+import logging
+import json
+import os
+import re
+import serial
```
kyleparrott: There'll be a pr on kubos-vagrant to fix this soon.

test/integration/linux/hello-world/script/hello_world.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
```
kyleparrott: They test 2 different cases: 
1) The compiled project tests uploading a built project executable to the board.
2) The script tests uploading a non-project file to the board. (Yes it's in a project directory for organization purposes, but it is still uploaded like any other project independent file)

test/integration/linux/test_runner.py
kyleparrott: the `kubos -t <target> flash` command does that for us without needing to explicitly set the target.

test/integration/linux/test_runner.py
```diff
@@ -0,0 +1,269 @@
...
+    def run_cmd(self, *args, **kwargs):
+        cwd = kwargs.get('cwd', os.getcwd())
+        save_output = kwargs.pop('save_output', False)
+        echo = kwargs.pop('echo', True)
+
+        if echo:
+            print ' '.join(args)
+        try:
+            if save_output:
+                return subprocess.check_output(args, **kwargs)
```
kyleparrott: If it causes things to fail, we shouldn't do that 😄 

`run_cmd` isn't called with `save_output=True` anywhere. And that makes sense as run_cmd returns and int, and `save_output=True` causes it to return a string. 

Other than fixing the format specifier to %i -> %d, which is my bad, I am a little confused as to what you want to change. (The flash command output is printed in the terminal as well without setting `save_output=True`)

CatFreed: Two different perspectives on this:
1. If save_output=True isn't specified anywhere, why is it in the function?
2. Having save_output=True available is nice for debugging. I turned it on in the run_cmd call from flash_project so that I could see the actual communication with the board.

So, if you want to have save_output for debugging, you need to change the way you're printing the output and return code from check_output.
If not, then take save_output out entirely.

test/integration/linux/generic_test_config.json
CatFreed: [system.destDir option](http://docs.kubos.co/sphinx/sdk-project-config.html#system)

test/integration/linux/test_runner.py
kyleparrott: The reason it's there is the `run_cmd` function is a duplicate of a [util function](https://github.com/kubostech/kubos/blob/master/tools/utils.py#L48) that is used in a number of our other tools scripts. 

Ideally once the final resting place for this script is determined I'll switch to importing that `run_cmd` versus just duplicating it like now.

One goal I have been trying to drive toward is less code duplication, and more of a common interface in a lot of the independent and "one-off" kubos tools and utility scripts. This is a side effect of that.

test/integration/linux/generic_test_config.json
```diff
@@ -0,0 +1,32 @@
...
+        "target": "kubos-linux-isis-gcc",
+        "timeout": 2
+    },
+    "login": {
+        "username": "root",
+        "password": "Kubos123"
+    },
+    "tests": [
+        {
+            "name": "Hello World Script Upload Test",
```
CatFreed: This test can't go first with the way things are now. When running completely from scratch, `kubos link -a` won't be run for the hello-world project, so no kubos-linux-isis-gcc target will be available. Additionally, there does need to be a hello-world/build/kubos-linux-isis-gcc folder in order for the flash to work.

test/integration/linux/hello-world/script/hello_world.sh
CatFreed: The permissions on this file need to be updated to be executable

test/integration/linux/hello-world/CONTRIBUTING.md
```diff
@@ -0,0 +1,4 @@
+Contributing to KubOS
```
plauche: I think we're good with the one contrib doc at the root of the repo. @davesims?

test/integration/linux/hello-world/script/hello_world.sh
plauche: :+1:

test/integration/linux/hello-world/README.md
```diff
@@ -0,0 +1,6 @@
+# KubOS Linux Hello World
+
+This is a simple application built on top of the [KubOS Linux Platform](https://github.com/kubostech/kubos-linux-build) 
+demonstrating the most baisc Hello World! functionality.
```
CatFreed: nit: spelling "basic"

kyleparrott:
Funny that exact function `add_test_failure`  was in an earlier version - it was removed for the hello-world failure abort scenario.. 😄

test/integration/linux/generic_test_config.json
```diff
@@ -0,0 +1,33 @@
...
+    "tests": [
+        {
+            "name": "Hello World Script Test",
+            "abort_on_failure": true,
+            "test_command" : "/home/system/usr/local/bin/hello_world.sh",
+            "flash_source": "./hello-world/script/hello_world.sh",
+            "expected_test_output": "Hello World!"
+        },
+        {
+            "name": "Hello world Compilation Test",
```
CatFreed: nit: Capitalize "World"
This test case should have abort_on_failure

test/integration/linux/readme.md
```diff
@@ -0,0 +1,74 @@
...
+* password - defaults to "Kubos123" if not provided
+
+This is not a required section. If it's not provided no login will be attempted.  It's assumed that your device is already logged into and has an active shell running.
+
+### Tests
+
+This section is an array of test descriptions. The tests are run in the order they are listed in the array.
+
+* name - The name of the test. For user recognition only. This has no effect on the test.
+* abort_on_failure = If true and this test fails, the runner will quit and not run any following tests. If not provided, it defaults to false.
```
CatFreed: nit: should have a hyphen separator, rather than an equals sign

test/integration/linux/generic_test_config.json
```diff
@@ -0,0 +1,34 @@
...
+            "abort_on_failure": true,
+            "test_command" : "/home/system/usr/bin/hello-world",
+            "build_source": "./hello-world",
+            "expected_test_output": "Hello World!"
+        },
+        {
+            "name": "C2 Ping Test",
+            "test_command" : "c2 core ping",
+            "expected_is_regex": true,
+            "expected_test_output": "Return Code: 0\nExecution Time \\d\\.\\d{6}\nOutput: Pong!"
```
plauche: One thought I had for a future version. Could we execute `$?` after the test command is run and just look at that return value? If we had a standardized set of return values it would shift more of the burden on the test programs to know the proper value, rather than maintaining text strings in two different places for each test program.

kyleparrott: Definitely that's a good idea. I think the next version of the script will be way easier to extend and add all kinds of new features to.

test/integration/linux/test_runner.py
```diff
@@ -0,0 +1,308 @@
...
+
+        if not os.path.isabs(proj):
+            proj = self.get_abs_path(start_dir, proj)
+
+        if os.path.isfile(proj):
+            # flash the project as a standalone file
+            flash_args.append(proj)
+            proj_dir = self.get_flash_project_dir()
+        elif os.path.isdir(proj):
+            # flash it like a regular kubos project - no additional args are needed
```
CatFreed: Super nit: Your comment formatting is not consistent. You've got "#stuff, #Stuff, # stuff".
Personally, I like "# Stuff" (space, capitalize first letter), but I'm not super particular as long as they're all the same

CatFreed: Running autopep8 will probably clean this up

test/integration/linux/hello-world/CONTRIBUTING.md
davesims: Yeah this can go.

test/integration/linux/hello-world/script/hello_world.sh
davesims: @kyleparrott `git update-index` probably [the best way](https://git-scm.com/docs/git-update-index) to set execute permissions.

/cc @CatFreed

PR #146, opened 2017-06-15
Telemetry QA tests

plauche:
Adding a test project to Kyle's integration test branch (#142).

This tests adding a telemetry subscriber and a telemetry publisher.

https://trello.com/c/0jkswLWo/38-telemetry-testing


test/integration/linux/telemetry-add-subscriber/source/main.c
```diff
@@ -0,0 +1,47 @@
...
+        .source.topic_id = QA_TEST_TOPIC,
+        .source.subsystem_id = 0,
+        .source.data_type = TELEMETRY_TYPE_INT,
+        .data.i = 100,
+        .timestamp = csp_get_ms()
+    };
+
+    socket_conn conn;
+
+    if (!telemetry_connect(&conn))
```
davesims: @plauche I think it would be good to print out success paths as well: "Telemetry connect succeeded", etc. What do you think?

test/integration/linux/generic_test_config.json
```diff
@@ -29,6 +29,18 @@
             "test_command" : "c2 core ping",
             "expected_is_regex": true,
             "expected_test_output": "Return Code: 0\nExecution Time \\d\\.\\d{6}\nOutput: Pong!"
+        },
+        {
+            "name": "Telemetry Add Publisher",
+            "test_command": "/home/system/usr/bin/telemetry-add-publisher",
```
CatFreed: FWIW, you should just be able to specify "telemetry-add-publisher", since the /home/system/usr/bin directory is part of the system's PATH

test/integration/linux/telemetry-add-publisher/source/main.c
```diff
@@ -0,0 +1,27 @@
...
+        .timestamp = csp_get_ms()
+    };
+
+    if (telemetry_publish(data))
+    {
+        printf("Telemetry publish success\n");
+    }
+    else
+    {
+        printf("Telemetry publish failed\n");
```
CatFreed: Nit: print the return code

test/integration/linux/telemetry-add-subscriber/source/main.c
```diff
@@ -0,0 +1,47 @@
...
+    {
+        printf("Telemetry connect failed\n");
```
CatFreed: Nit: print return codes with all error messages

test/integration/linux/telemetry-add-subscriber/source/main.c
```diff
@@ -0,0 +1,47 @@
+#include <telemetry/telemetry.h>
+#include <csp/arch/csp_time.h>
+#include <stdio.h>
+
+#define QA_TEST_TOPIC 101
```
CatFreed: Unless I'm missing something, 101 is the message code for supervisor_uptime, which is an iOBC specific message. Ideally we want this test to be portable, so we'd need to use a common publisher.

I'm pretty sure there isn't a common publisher right now, so I think we have two options:
1. Create a common publisher (probably less than ideal)
2. Remove this test from the generic test configuration and only have it in the iobc test configuration (which has not been created yet, but should just contain all the test cases we're creating right now) and make a note to update this test to make it portable in the future.

davesims:
[Telemetry Testing](https://trello.com/c/0jkswLWo/38-telemetry-testing)

test/integration/linux/telemetry-add-publisher/source/main.c
plauche: Return value in this case is true/false

test/integration/linux/telemetry-add-subscriber/source/main.c
plauche: Currently almost all of the telemetry functions return true/false rather than a return code.

test/integration/linux/telemetry-add-subscriber/source/main.c
plauche: Hmmm for now we can make this test iOBC specific, then remake it in 1.1 to be more generic.

test/integration/linux/telemetry-add-subscriber/source/main.c
CatFreed: Ah, ok. Do you have any concerns about how easy or hard it will be to debug telemetry problems going forward?

plauche: I think it is worth exploring expanding the telemetry return codes in the near future (1.1).

test/integration/linux/isis-obc/iobc-test-config.json
```diff
@@ -0,0 +1,58 @@
...
+        "target": "kubos-linux-isis-gcc",
+        "timeout": 15
+    },
+    "login": {
+        "username": "root",
+        "password": "Kubos123"
+    },
+    "tests": [
+        {
+            "name": "Hello World Script Test",
```
kyleparrott: Do we want to combine the "generic" tests into the iobc specific tests?

CatFreed:
You need to rebase this onto the master branch, but then it should be good to go

PR #148, opened 2017-06-19
Tweaking sz/rz commands to allow file transfer resume

CatFreed:
ZModem transfer is very slow, causing kpack files to take ~30 minutes to be sent. It is not uncommon for the transfer to timeout. Rather than start back over, it would be much nicer to simply resume the previous transfer.


CatFreed:
[OS Update Testing](https://trello.com/c/72s0D6o1/40-os-update-testing)

CatFreed:
@plauche It's not great behavior...
- If you send literally the same file, nothing actually gets sent and so the file transfer completes successfully.
- If you send a file of the same name which is larger than the original, it will transfer starting at len(original) until the end of the larger file.
- If you send a file of the same name which is smaller than the original, the new file will get transferred on top of the old file. So the old file size stays, but the data is now wrong...

However, the actual resume mechanism works great :stuck_out_tongue:

targets/target-kubos-linux-isis-gcc/minicom/flash.sh
```diff
@@ -83,8 +83,10 @@ expect {
 timeout 3600
 send "mkdir -p $2"
 send "cd $2"
+if ${is_upgrade} = 1 goto check_space
```
davesims: Out of curiosity, where does `is_upgrade` come from? I see it's documented as 'global' above, is it set from an env somewhere?

CatFreed: Line 281. If we're flashing an *.itb file, we assume it's an upgrade.

plauche:
I'm concerned about the effect this change will have on normal application development.  Right now the script appears to always resume transmissions, correct? We need the ability to overwrite files on the iobc when working on user applications.

Would it be possible to detect if the last flash failed and attempt a resume if the same file is sent again?

CatFreed:
@plauche Good point. I'd really only been thinking about resuming the upgrade transfers. I think for right now I'll just update the script so that only upgrade files use the "resume" option. Going forward, I'd like to add an option to the flash command to allow users to specify that they'd like to resume a previous flash, but that will be post-1.0.

PR #150, opened 2017-06-20
C2 Supervisor Request Retry

kyleparrott:
This pr implements retrying a supervisor command up to a set number of times if previous responses came back with bad checksums. If it fails to get a valid response each time then it should fail much more gracefully than before.

It also cleans up a couple of compiler warnings, formatting issues and should align it a little more closely with our standards doc.


kyleparrott:
[C2 output on bad supervisor output](https://trello.com/c/BOUmRqLD/55-c2-output-on-bad-supervisor-output)

commands/source/core.c
```diff
@@ -71,61 +49,82 @@ int ping()
...
     {
-        printf("There was an error getting the supervisor version information. Error: %i\n", result);
-        return result;
+        result = supervisor_get_version(&version);
+        if (result)
+        {
+            printf("iOBC Supervisor Version: %c.%c.%c\n", version.fields.major_version, version.fields.minor_version, version.fields.patch_version);
+            return NO_ERR;
+        }
+        retries++;
```
CatFreed: You could just use a for loop for this, instead of manually incrementing the counter

commands/source/parser.c
```diff
@@ -90,6 +90,7 @@ bool core_parse_args(int argc, char ** argv, char * cmd_string)
     return true;
 }
 
+
```
CatFreed: nit: extra line

PR #153, opened 2017-06-21
Doc changes per isis review

CatFreed:
Making minor doc changes that Gerard has suggested


docs/working-with-the-iobc.rst
```diff
@@ -754,6 +754,8 @@ SPI
 The external SPI bus is not currently available to the userspace. It
 will be added in a future release.
 
+.. _user-accounts:
+
 User Accounts
```
jacoffey3: When should they create a user account? Is that a necessary step (necessary as in we want them to do it?)

CatFreed: It's not a necessary step. We include the kubos/Kubos123 user account for this purpose.

A recommended step would be changing the root password, though...

CatFreed:
[ISIS Review - Doc Fixes](https://trello.com/c/Awm7YA4A/27-isis-review-doc-fixes)

docs/working-with-the-iobc.rst
jacoffey3: Whoops. Forgot to check back in. Sorry.

PR #147, opened 2017-06-18
Added initial config for ipc

TheHollidayInn:
I believe I am on the right track with this, however, I am getting this error:

screen shot 2017-06-18 at 2 18 46 pm

Can anyone direct me to where I am making the mistake? Thanks!

PR #161, opened 2017-06-27
Moving IPC docs to sphinx

plauche:
CircleCi was being a pig and not running the original PR (#147) by @thehollidayinn, so I had to create this internal branch.


plauche:
Ok nvm the original PR was finally picked up.

PR #167, opened 2017-06-30
Adding beaglebone target

kyleparrott:
@CatGarab  My bad on forgetting this!! These are the only remaining changes I have for the BBB.

This is the target I have been using to build for the BBB. The only real difference from the iOBC target is the 🔧  ⛓  (toolchain).


CatGarab:
Cool :) The target.json and flashing scripts are identical between the iobc and the beaglebone. Could a parent target be used for this? It'd probably be a new generic linux target

kyleparrott:
Yep that should work just fine.

CatGarab:
Ok. Do you want to add the target now, or are you thinking of just merging this PR now and then it'll get added/updated later?

kyleparrott:
I'll leave it up to you. Either way is fine by me. 

There's some more changes to our CI configs so  the beaglebone will be built as well that will be in a later PR.

CatGarab:
If it's something that'll be fast (<30 min), i'd say do it now. Otherwise, just merge in the master changes needed, and we'll go ahead and get this thing merged

kyleparrott:
I'm heads down on MT at the moment. It would take me a little longer to circle around and get the target hierarchy setup and tested properly so I am going to merge this to master once the checks pass.

If this drags out or gets in your way I can pick it back up sooner.

targets/target-kubos-linux-beaglebone-gcc/minicom/flash.sh
```diff
@@ -0,0 +1,272 @@
...
+if ${is_app} = 1 send "rm -f /home/system/etc/init.d/S*${name}"
+send "rz -w 8192"
+! sz -w 8192 ${path}
+if ${is_upgrade} = 1 send "fw_setenv kubos_updatefile ${name}"
+if ${is_run} = 0 goto exit
+send "start-stop-daemon -K -v -p /var/run/${name}.pid"
+send "start-stop-daemon -S -v -m -b -p /var/run/${name}.pid --exec $2/${name}"
+exit:
+send "exit"
+end:
```
davesims: Nit: Would some output on error condition be appropriate here?

targets/target-kubos-linux-beaglebone-gcc/minicom/flash.sh
```diff
@@ -0,0 +1,272 @@
+#!/bin/bash
```
plauche: Are there any differences between this and the iobc flash script? If they are the same then at some point we might consider pulling them into a higher level, generic kubos-linux target.

kyleparrott: Yep, it is identical. This PR was to get everyone caught up and all my BBB changes pushed up. A new higher level linux target containing that and a huge amount of the shared details in their target.jsons will be created in a future PR.

This is tracked in [trello](https://trello.com/c/6PSfiKPR/30-create-higher-level-iobc-bbb-shared-linux-target)

PR #168, opened 2017-07-10
[SPIKE] Event-Driven Architecture

plauche:
This PR represents an initial code sketch of an event driven architecture. It is intended to start the conversation, help expose gaps where further R&D is required, and provide a possible structure to start building on.

## Rough Layout of Components
```                                                                                                                                                                                                                                                                   
+-------------------------------------+                      +-------------------------------------+                      +-------------------------------------+ 
|                                     | Event Subscription   |                                     |   Event Request      |                                     | 
|                                     |--------------------> |                                     |--------------------->|                                     | 
|          User Application           |                      |            Event Broker             |                      |             EPS Service             | 
|                                     |                      |                                     |                      |                                     | 
|                                     |<-------------------- |                                     |<---------------------|                                     | 
+-------------------------------------+    Event Result      +-------------------------------------+  Event Registration  +-------------------------------------+ 
                                           Publishing                                                 Event Firing                           |                    
                                                                                                                                             |                    
                                                                                                                                             |                    
                                                                                                                                             | EPS Interface Calls
                                                                                                                                             |                    
                                                                                                                                             v                    
                                                                                                                          +-------------------------------------+ 
                                                                                                                          |                                     | 
                                                                                                                          |                                     | 
                                                                                                                          |              EPS API                | 
                                                                                                                          |                                     | 
                                                                                                                          |                                     | 
                                                                                                                          +-------------------------------------+ 
```

## Rough Sequence Diagram
```
+-------------------------------------+              +-------------------------------------+               +-------------------------------------+              +-------------------------------------+
|                                     |              |                                     |               |                                     |              |                                     |
|                                     |              |                                     |               |                                     |              |                                     |
|          User Application           |              |            Event Broker             |               |             EPS Service             |              |              EPS API                |
|                                     |              |                                     |               |                                     |              |                                     |
|                                     |              |                                     |               |                                     |              |                                     |
+------------------+------------------+              +------------------+------------------+               +------------------+------------------+              +------------------+------------------+
                   |                                                    |                                                     |                                                    |
                   |                                                    | <---------------------------------------------------+                                                    |
                   |                                                    |      register_event(ON_POWER_READING)               |                                                    |
                   |                                                    |                                                     |                                                    |
                   +--------------------------------------------------> |                                                     |                                                    |
                   |                onPowerReading                      |                                                     |                                                    |
                   |                                                    +---------------------------------------------------> |                                                    |
                   |                                                    |      request_event(ON_POWER_READING)                |                                                    |
                   |                                                    |                                                     |                                                    |
                   |                                                    |                                                     +---------------------+                              |
                   |                                                    |                                                     |                     |                              |
                   |                                                    |                                                     |                     |                              |
                   |                                                    |                                                     |                     |                              |
                   |                                                    |                                                     |  EPS Power Level    |                              |
                   |                                                    |                                                     |  Monitoring Thread  +----------------------------> |
                   |                                                    |                                                     |                     |   eps_get_power_level        |
                   |                                                    |                                                     |                     |                              |
                   |                                                    |                                                     |                     |                              |
                   |                                                    |                                                     |                     |                              |
                   |                                                    |                                                     +---------------------+                              |
                   |                                                    |                                                     |                                                    |
                   |                                                    |                                                     |                                                    |
                   |                                                    | <---------------------------------------------------+                                                    |
                   |                                                    |      fire_event(ON_POWER_READING, data)             |                                                    |
                   | <--------------------------------------------------+                                                     |                                                    |
                   |        publish_event(ON_POWER_READING, data)       |                                                     |                                                    |
                   |                                                    |                                                     |                                                    |
                   |                                                    |                                                     |                                                    |
                   +-------------------+                                |                                                     |                                                    |
                   |                   |                                |                                                     |                                                    |
                   |                   |                                |                                                     |                                                    |
                   | runOnPowerReading |                                |                                                     |                                                    |
                   | callback for      |                                |                                                     |                                                    |
                   | event receive     |                                |                                                     |                                                    |
                   |                   |                                |                                                     |                                                    |
                   |                   |                                |                                                     |                                                    |
                   +-------------------+                                |                                                     |                                                    |
                   |                                                    |                                                     |                                                    |
                   +                                                    +                                                     +                                                    +
```

## Application detail
```
+----------------------------------------------------------------------------------------------------------+                                           
|                                                 Event Driven Application                                 |                                           
|                                                                                                          |                                           
|                                                                                                          |                                           
|                                                                                                          |                                           
|                                                                                                          |                                           
|                                                                                                          |                                           
|         +--------------------------------+                      +------------------------------------+   |                                           
|         |                                |                      |                                    |   |                                           
|         |                                |                      |          Main (Loop)               |   |                                           
|         |  runOnPowerReading             |                      |                                    |   |                                           
|         |  (call back for power reading) |                      |  start_event_loop()                |   |                                           
|         |                                |                      |  onPowerReading(runOnPowerReading) |---|-------------------------------------->    
|         |                                |                      |                                    |   |     request_event(ON_POWER_READING)       
|         +--------------------------------+                      +------------------------------------+   |                                           
|                          ^                                                         |                     |                                           
|                          |                                                         |start_event_loop     |                                           
|                          |                                                         |spawns thread        |                                           
|                          |                                                         |                     |                                           
|                          |                                                         v                     |                                           
|         +--------------------------------+                      +------------------------------------+   |                                           
|         |     Event  Fire Thread         |                      |       Event Listening Thread       |   |                                           
|         |                                |                      |                                    |   |                                           
|         |                                |                      |                                    |   |                                           
|         |                                |<---------------------|   Event receive/callback trigger   |<------------------------------------------    
|         |                                |Spawn one-shot thread |                                    |   |      receive_event(ON_POWER_READING)      
|         |                                |for event callback    |                                    |   |                                           
|         |                                |                      |                                    |   |                                           
|         +--------------------------------+                      +------------------------------------+   |                                           
|                                                                                                          |                                           
|                                                                                                          |                                           
+----------------------------------------------------------------------------------------------------------+                                           
```

## Service Details
```
+--------------------------------------------------------------------------------+                                                
|                          Event Driven Service                                  |                                                
|                                                                                |                                                
|                                                                                |                                                
|                                                                                |                                                
|                                                                                |                                                
|                                                                                |                                                
|        +-------------------------------+             +-----------------------+ |                                                
|        |  Power Level Reading Thread   |             |        Main Loop      | |                                                
|        |                               |             |                       | |    register_event(ON_POWER_READING)            
|        |                               |             |                       |------------------------------------------>       
|        |  eps_get_power_level()        |             |                       | |                                                
|        |                               |             |                       | |                                                
|        +-------------------------------+             |                       | |                                                
|                        |                             |                       | |                                                
|                        |                             |                       | |                                                
|                        v                             |                       | |                                                
|              +-------------------+                   |                       |<-----------------------------------------        
|              | Power Level Queue |                   |                       | |    request_event(ON_POWER_READING)             
|              +-------------------+                   |                       | |                                                
|                        |                             +-----------------------+ |                                                
|                        |                                                       |                                                
|                        v                                                       |                                                
|        +-------------------------------+                                       |                                                
|        | Power Level Processing Thread |                                       |                                                
|        |                               |                                       |                                                
|        |                               |                                       |                                                
|        |                               |--------------------------------------------------------------------------------------->
|        |  check power reading requests |   fire_event(ON_POWER_READING)        |                                                
|        |                               |                                       |                                                
|        |                               |                                       |                                                
|        +-------------------------------+                                       |                                                
|                                                                                |                                                
|                                                                                |                                                
+--------------------------------------------------------------------------------+                                                
```


plauche:
[SPIKE: Event Architecture R&D](https://trello.com/c/68WFghFJ/26-spike-event-architecture-rd)

eps/eps-service/source/service.c
```diff
@@ -0,0 +1,111 @@
...
+ *   - Determine if parameters are met
+ *   - Send back message if conditions are good
+ * - Thread should exit on service exit
+ */
+void power_level_process_thread(void)
+{
+    uint32_t power_level;
+    while (1)
+    {       
+        dequeue_power_level(&power_level);
```
CatGarab: Is this where an event would be queued for the fetch_event function to process later?

events-broker/source/broker.c
```diff
@@ -0,0 +1,75 @@
...
+        /**
+         * Event Publish
+         * Source -> Broker -> Client
+         *
+         * - Locate connection with app_key
+         * - Sends event_key and data back to app
+         */
+        {
+            event_pub_t pub;
+            send_event_data(pub.app_key, pub.event_key, pub.data);
```
CatGarab: Is this the same as "publish_event" from your sequence diagram?

events-broker/source/broker.c
```diff
@@ -0,0 +1,75 @@
...
+         * Event Registration
+         * Source -> Broker
+         *
+         * - Adds source_key and event_key to list of 
+         *   available events
+         * - Stores and associates connection with source_key
+         */
+        {
+            event_reg_t reg;
+            register_event(reg.source_key, reg.event_key);
```
CatGarab: Based on your diagrams, it seems like the declaration of register_event should belong here-ish, instead of just a call to it, unless you're intending for this to be how custom user-events are added.

examples/kubos-eps-example/source/main.c
```diff
@@ -0,0 +1,51 @@
...
+    
+    // decode_power_reading(buffer, &power_reading);
+    printf("Received power reading %d\n", power_reading);
+}
+
+
+int main(void)
+{
+    printf("Starting EPS Event Example\n");
+
```
CatGarab: Your application details diagram has a `start_event_loop` call before the first event request

kubos-events/source/kubos-events.c
```diff
@@ -0,0 +1,81 @@
...
+     *   - Request should have event_key and buffer
+     * - Call add_event_listener with event_key and cb
+     */
+}
+
+void start_event_loop(void)
+{
+    /**
+     * - Spawn off event_loop_thread
+     */
```
CatGarab: Assuming this listens for whatever `send_event_data` does?

kubos-events/source/kubos-events.c
```diff
@@ -0,0 +1,81 @@
...
+     *   list[event_key] = cb;
+     */
+}
+
+void register_event(char * source_key, char * event_key)
+{
+    /**
+     * - Creates and sends event registration message to event broker
+         - Needs to send source_key and event_key pair
+    */
```
CatGarab: Seems like this should be something like just adding an entry to a table, rather than sending a message for another function to pick up

examples/kubos-eps-example/source/main.c
```diff
@@ -0,0 +1,51 @@
...
+    onPowerLevel(30, &runOnLowPower);
```
davesims: For clarity I think this should be called something like `registerPowerLevelListener`, and the `runOnPowerReading` method above should be `onPowerReading`. Typically the '`onFizzBuzz`' naming convention in an evented system indicates the code that will be run when the event is fired.

eps/eps-service/source/service.c
plauche: This loop kinda works on the back end of the result of `fetch_event`. Event requests from `fetch_event`  are stored and examined here with each queued up power level reading. If the reading triggers one of the events then it is fired back to the event broker.

events-broker/source/broker.c
plauche: Yup that is correct. Terms are not entirely consistent at the moment :)

kubos-events/source/kubos-events.c
plauche: My naming here is a bit confusing. This `register_event` is intended to create a message and send it to the `events_broker` which is a separate process. The `register_event` inside of the `events_broker` will be creating an entry in a table. One or both of these functions should be renamed.

examples/kubos-eps-example/source/main.c
plauche: :+1:

davesims: @plauche On second thought I take it back, this is also a pretty typical convention. We're saying OnEvent -> run this function. Not sure why I didn't read it that way. This actually looks great.

kyleparrott:
This is a fantastic start @plauche 🎉  You're code is really clean and easy to understand.

My initial thoughts are about the queues:

1) According to your diagrams it looks like each event type would have it's own queue?  That seems like a lot. I think with some research we could come up with a way to limit the number of queues and threads reading and writing to those queues.

2) The bigger issue in my mind is concurrency control with a large number of queues. This seems like we could quickly get into concurrency/threading hell with reading and processing threads per queue concurrently updating and reading from their respective queue.

davesims:
@plauche 💯 ✖️ :bow: on this, I think we're heading in the right direction for sure. Very, very clean code. Even my weak c-fu could follow. :bowtie: 

I'm wondering if we couldn't reduce the thread count (and one _always_ should reduce the thread count) with a single `EventLoop` that works off events from a `Hash`? We'd have to implement a `Hash` or [pull one in](http://freecode.com/projects/libcfu), but I could see something along the lines of a single loop that pulls events off of a single queue, and each `Event` struct has a key, which the loop then looks up in the subscriber `Hash`, which stores an array of callback pointers that are then worked off in order. 

If that could work, it reduces the total concurrency greatly: a single event worker taking each callback in turn guarantees that, for instance, multiple threads aren't contending for the radio at once.

eps/eps-events/source/events.c
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * - Performs registration for specified event + encoded data parameter
+ * - Sends registration message to message broker
+ *   - Spawns thread waiting on event response
+ */
+void onPowerLevel(uint32_t power_level, event_callback callback)
+{
+    uint8_t data_buffer[100];
+
+    request_event(EVENT_EPS_ON_POWER_LEVEL, callback, data_buffer);
```
davesims: Niggle: to me `register_handler` or the like is more descriptive than `request_event`.

kubos-events/source/kubos-events.c
davesims: So @plauche `register_event` comes from a background service that monitors the peripheral, and when something happens puts an event on a queue that the `EventBroker` owns and reads off of, locates the subscribers and publishes to them?

davesims:
One request @plauche: can we move the asciiflows into a /docs folder somewhere for easier viewing?

eps/eps-service/source/service.c
CatGarab: Hmm, ok. So going forward there would be some sort of check? 
```
if(someone wants power)
{
     dequeue_power_level();
     tell_fetcher(level)
}
```

plauche:
This spike has ended. Current production work is taking place in #188.

PR #171, opened 2017-07-21
Add C unit tests to Circle builds

plauche:
We have a few unit tests created but we don't currently build and run them on each build. 

This PR will...
- [x] Make Circle detect, build and run any unit tests for kubos modules
  - Note - This is natively (x86-linux-native) building and running
- [x] Provide a way to detect which projects *should* be built natively vs cross-compiled
- [x] Fix any code issues uncovered when running all unit tests


tools/build.py
```diff
@@ -129,24 +178,38 @@ def main():
...
     elif args.list_modules:
         ret = builder.list_modules()
+    elif args.list_testable_modules:
+        ret = builder.list_testable_modules()
     elif args.list_changed_modules:
         ret = builder.list_changed_modules(args.list_changed_modules)
+    elif args.test:
+        ret = builder.test(module_name=args.module, target_name='x86-linux-native')
+    elif args.all_tests:
+        ret = builder.test_all_modules(target_name='x86-linux-native')
```
CatGarab: You have one test with the STM32F4. Should that be included here?

tools/build.py
```diff
@@ -113,6 +153,11 @@ def build_all_combinations(self):
                     ret = build_ret
         return ret
 
+def get_branch_name():
+    branch_key = 'CIRCLE_BRANCH'
+    default_branch = 'master'
+    return os.environ[branch_key] if branch_key in os.environ else default_branch
```
kyleparrott: I think we should print some kind of warning if the `CIRCLE_BRANCH` env isn't set, that would be some weird kind of issue if that wasn't set in a circle build.

tools/build.py
```diff
@@ -129,24 +178,38 @@ def main():
...
 
     args = parser.parse_args()
 
     builder = KubosBuilder()
 
-    ret = 0
+    ret = 0 
+
+
+    print("Running kubos update..")
```
CatGarab: nit: ellipses have 3 periods

services/telemetry/telemetry-rt/module.json
```diff
@@ -13,5 +13,8 @@
     },
     "testDependencies": {
         "cmocka": "kubostech/cmocka"
-    }
+    },
+    "testTargets": [
+        "stm32f407-disco-gcc"
```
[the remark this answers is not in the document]
plauche: We currently aren't testing against the MSP430

telemetry-storage/module.json
```diff
@@ -11,6 +11,9 @@
         "kubos-rt":"kubostech/kubos-rt"
       }
     },
+    "testTargets": [
+        "x86-linux-native"
```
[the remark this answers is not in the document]
plauche: Probably...but for now I'm going to leave it off because we currently aren't really testing or using kubos-rt.

tools/build.py
plauche: I haven't quite worked out how to do target-based testing with this script. The main focus for this PR is getting native unit tests running under Circle.

tools/build.py
```diff
@@ -129,24 +178,41 @@ def main():
...
                         help='Lists all modules found')
-    parser.add_argument('--list-changed-modules', action="store", nargs="?", 
+    parser.add_argument('--list-testable-modules', action='store_true', default=False,
+                        help='Lists all modules with defined tests')
+    parser.add_argument('--list-changed-modules', action="store", nargs="?",
                         dest="list_changed_modules", const="HEAD^!",
                         help='Lists modules that have changed. By default will diff against '
                         'the last commit. The git diff path desired can also be passed in')
-        
+
```
CatGarab: nit: extra line

tools/build.py
```diff
@@ -129,24 +178,41 @@ def main():
...
+                        help='Builds against local source')
 
     args = parser.parse_args()
 
     builder = KubosBuilder()
 
-    ret = 0
+    ret = 0 
+
+
```
CatGarab: nit: extra line

PR #173, opened 2017-07-26
Documentation for Beaglebone Black Target

CatGarab:
Adding docs for KubOS Linux on the Beaglebone Black target.

I decided that repeating large chunks of information in multiple places was less than desirable, so there is now a common "Using KubOS Linux" doc for working with Kubos projects in the linux userspace.


CatGarab:
[Beaglebone Black Target](https://trello.com/c/S9OH2xI9/46-beaglebone-black-target)

PR #177, opened 2017-07-28
Removing SPI0 from BBB. Adding small doc updates

CatGarab:
Updating documentation to remove the SPI0 bus from the Beaglebone Black documentation. It was removed from the board because the pins conflict with I2C1.

Also adding some small notes that have been in my head.

### DIFF
--- a/docs/kubos-linux-on-bbb.rst
+++ b/docs/kubos-linux-on-bbb.rst
@@ -243,11 +243,17 @@ The KubOS Linux installation process is composed of two high-level steps:
 To perform a full default installation, two files are needed:
 
   - A KubOS Linux SD card image
-  - An aux_sd1 image
+  - An aux_sd image
   
 All of these files can be obtained from `our KubOS Linux Releases page on GitHub <https://github.com/kubostech/kubos-linux-build/releases>`__
 
 Download the latest `KubOS_Linux.zip` file and then unzip the files for the Beaglebone Black. They're located in the `KubOS_Linux/{version}/Beaglebone-Black` folder.
+
+.. note::
+
+    The Beaglebone Black can also be used as a development board for the Pumpkin MBM2.
+    If you would like to use it for this purpose, please follow the
+    :ref:`installation instructions for the MBM2 <installation-process-mbm2>` instead.
 
 Pre-Requisites
 ~~~~~~~~~~~~~~
@@ -355,6 +361,10 @@ Now flash the micro SD card with the auxiliary SD card image. This image contain
 KubOS Linux upgrade partition and the second user data partition.
 
 Once the flash process has completed, put the card back into the microSD slot.
+
+.. warning::
+
+    If you do not have a microSD card in the board, the system will not boot.
 
 The installation process is now complete.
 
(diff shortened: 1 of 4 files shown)


CatGarab:
[Test everything](https://trello.com/c/pyX2oeqg/50-test-everything)

docs/working-with-the-bbb.rst
```diff
@@ -45,11 +45,12 @@ with peripheral devices. Currently, users should interact with these
...
 
-.. warning::
-
-    Some pins are re-used by multiple buses. As a result, you will not be
-    able to use all buses simultaneously.
+.. note::
 
+    KubOS Linux for the Pumpkin MBM2 can be used instead of KubOS Linux
+    for the Beaglebone Black. In this case, some peripherals won't be
+    available. See the :ref:`peripherals-mbm2` section for more information.
```
jacoffey3: What do you mean here? Wouldn't more peripherals be available on the MBM2 rather than less?

docs/kubos-linux-on-bbb.rst
```diff
@@ -356,6 +362,10 @@ KubOS Linux upgrade partition and the second user data partition.
 
 Once the flash process has completed, put the card back into the microSD slot.
 
+.. warning::
+
+    If you do not have a microSD card in the board, the system will not boot.
```
jacoffey3: Is this always true? What blocks it from booting?

docs/working-with-the-bbb.rst
jacoffey3: Okay so you mean pin access, not endpoint devices?

docs/kubos-linux-on-bbb.rst
CatGarab: Right now, yes. By default the kernel will attempt to load the rootfs from the eMMC with device /dev/mmcblk1p2. When the sd card is plugged in, the sd card is mmcblk0 and the eMMC is mmcblk**1**. When the sd card is not plugged, the eMMC becomes mmcblk**0**. The kernel tries to load the rootfs from the same device, which now doesn't exist, so the system fails to boot.

This is absolutely something that should be changed, but for the sake of time I left it like this for now so I could get the rest of the system up and running.

jacoffey3: Got it. Can you put a card up in the backlog for it? Just so we don't forget.

docs/working-with-the-bbb.rst
CatGarab: Yeah

docs/kubos-linux-on-bbb.rst
CatGarab: [It's already there](https://trello.com/c/QarGCPuH/44-bbb-handle-only-emmc)

CatGarab: [This one is also kind of related](https://trello.com/c/6VMREd15/37-pumpkin-auto-dynamic-boot)

jacoffey3: Should've known lol

PR #185, opened 2017-08-09
Fix broken links in first-rt-project.rst

PBeninga:
Fix broken links for example projects in documentation


plauche:
@PBeninga as per our talk at SmallSat, this PR is already covered by previous work so I'll close it.

PR #186, opened 2017-08-15
Added 'none' board type for simple powerup/powerdown commands

jessehamner:
Some light editing of a few descriptions, plus glue code to enable the 'NoBoard' Target type.
Added some new attributes for each pin, given the potential need for pins to be set to input or output.


test/integration/jenkinsnode/cistack/none.py
```diff
@@ -0,0 +1,27 @@
+import os
```
plauche: I'd suggest we pick a name which isn't so close to the Python keyword `None`.

test/integration/jenkinsnode/ci.py
```diff
@@ -47,6 +47,10 @@ def main():
     logging.debug("Script started %s." % str(NOW))
     logging.info("Arguments have been parsed.")
 
+#    sofile=open('/tmp/stdoutfile.txt', 'w')
```
plauche: Should this code by commented/removed or triggered by a flag?

test/integration/jenkinsnode/ci.py
```diff
@@ -66,6 +70,31 @@ def main():
...
+
+# set up the GPIO pins as requested
+    target.setupboard()
+    sleep(1)
+
+# create a dict of Pin objects:
+    pins = target.getpins()
+    logging.debug(str(pins))
+
+    if args.powerup is True:
```
plauche: What is the difference between the `args.powerup` flag and `args.command` with a value of `"powerup"`? They appear to be different flags with identical functionality.

jessehamner: Old habits die hard, and I always forget exact keywords. I'll pull out the 'command' functions.

test/integration/jenkinsnode/ci.py
jessehamner: Ah, that should be removed. Thanks for catching it.

test/integration/jenkinsnode/cistack/none.py
jessehamner: Very fair. I thought of that but wanted to throw the PR up quickly -- the iOBC is benefiting from this functionality and I need to hurry up.

PR #187, opened 2017-08-17
Adding iOBC UART doc + demo

CatGarab:
Adding doc blurb about the iOBC's UART ports

Also adding UART TX/RX demo program for KubOS Linux

It's not well fleshed out, but it does work, so I want to save this code somewhere accessible for future reference.


CatGarab:
[iOBC UART communication](https://trello.com/c/2R6R3xOA/61-iobc-uart-communication)

plauche:
Do these programs belong in the `examples` folder, rather than creating a new `demos` folder?

CatGarab:
@plauche Ultimately they could go there, but right now they're rough enough that I don't really want other people to use them. I'd like to come back later and flesh them out with what the options are and how exactly `termio` works in the context of UARTs.

CatGarab:
Cleaned the examples up to a point where I'm happy with them. They have now been migrated to the examples folder

PR #189, opened 2017-08-23
Documenting and testing iOBC FRAM

CatGarab:
The iOBC's FRAM will be exposed in the licensed version of KubOS Linux. Documenting the peripheral and how to use it.

Adding a new QA test `iobc-fram-test` to the iOBC's QA suite.


CatGarab:
[IOBC FRAM](https://trello.com/c/6CtVOg1d/19-iobc-fram)

docs/working-with-the-iobc.rst
```diff
@@ -217,6 +217,59 @@ SPI
...
+~~~~
+
+.. note:: This peripheral is only available with the licensed version of KubOS Linux
+    for the iOBC, available from ISIS.
+    
+The iOBC has 192 KBytes of FRAM storage available. It can be accessed via
+the ``/sys/class/spi_master/spi0/spi0.0/eeprom`` device with the standard
+open/read/write/close functions.
+
+.. note:: The device name says "eeprom" rather than "fram" because both device types
```
jacoffey3: Can we not change this? Seems like a simple thing to reduce confusion in future code.

docs/working-with-the-iobc.rst
```diff
@@ -217,6 +217,59 @@ SPI
...
 
+FRAM
+~~~~
+
+.. note:: This peripheral is only available with the licensed version of KubOS Linux
+    for the iOBC, available from ISIS.
+    
+The iOBC has 192 KBytes of FRAM storage available. It can be accessed via
+the ``/sys/class/spi_master/spi0/spi0.0/eeprom`` device with the standard
+open/read/write/close functions.
```
jacoffey3: I think you need to go a bit more in depth on how to actually use it. Your example addresses a basic use case, but what memory addresses do I actually write to? You use the example address of 0x100, but how many other addresses are there? If you write 10 bytes to 0x100, and then write a byte to 0x101, would you overwrite some of your previous information? I'm sure there is a guide on this that already exists somewhere, but we need to link to it at a minimum.

docs/working-with-the-iobc.rst
CatGarab: We can, it just requires 1 of 2 things:
- patch the kernel (there is an FRAM patch out there already, so I wouldn't have to actually write one)
- see if we can create a more user friendly device name like ``/dev/fram``. If I'm right, it's fairly straight forward and would probably be done by an init script at startup. If I'm wrong, I'm not sure how much digging it'll take to get it to work.

docs/working-with-the-iobc.rst
CatGarab: I tweaked the wording a bit, which should clarify things. I didn't mention the 0x100 vs 0x101 thing because I'm now using the word "offset" and I think it should be a pretty straight-forward conclusion that writing to any offset will overwrite whatever was previously there.

test/integration/linux/iobc-fram-test/source/main.c
```diff
@@ -0,0 +1,103 @@
...
+#include <string.h>
+#include <unistd.h>
+
+int main(int argc, char * argv[])
+{
+
+    char * device     = "/sys/class/spi_master/spi0/spi0.0/eeprom";
+    char * teststring = "FRAM Test String";
+
+    char buf[sizeof(teststring)] = { 0 };
```
plauche: I think in this case you'll want to use `strlen` to get the length of the string, not `sizeof` which will just give the size of a pointer.

test/integration/linux/iobc-fram-test/module.json
```diff
@@ -0,0 +1,15 @@
...
+    "bin":"./source",
+    "license":"Apache-2.0",
+    "name":"iobc-fram-test",
+    "repository":{
+        "url":"git://<repository_url>",
+        "type":"git"
+    },
+    "version":"0.1.0",
+    "dependencies":{
+        "csp":"kubostech/libcsp#~1.5.0"
```
plauche: Does this need the `csp` dependency?

test/integration/linux/iobc-fram-test/source/main.c
CatGarab: Yup :/ whoops

CatGarab: Decided I do still like sizeof, since it's a compile-time function, but I'm doing it correctly now :P

ISISGerard:
Probably a stupid question, but where can I actually read the FRAM-related documentation without looking at the syntax?

docs/working-with-the-iobc.rst
ISISGerard: I'm also a little concerned about the line that specifies the SPI device number and the fact that that's public...

CatGarab:
This PR is being moved into kubostech/iobc-klb#5

PR #194, opened 2017-09-26
Generic radio API

CatGarab:
First pass at creating a generic radio API. Modeling after the generic EPS API that's in progress. This will be used in conjunction with the ISIS TRXVU radio's API (which is a WIP).

In theory, specific radio APIs will go in the `radio` folder, giving a final structure like this:

kubos/radio/
  - radio-api: generic API
  - example1-radio-api: specific radio API 1
  - example2-radio-api: specific radio API 2

A new documentation section has been created for this (and the other future device APIs, like the EPS) called "Device Interfaces". It's basically what Kubos Core was supposed to be.


CatGarab:
[Radio API](https://trello.com/c/B7k4uhPR/29-radio-api)

radio/radio-api/radio-api/radio.h
```diff
@@ -0,0 +1,117 @@
...
+ * @param [in] radio_config Pointer to the radio configuration structure
+ * @return KRadioStatus RADIO_OK if OK, error otherwise
+ */
+KRadioStatus k_radio_configure(uint8_t * radio_config);
+/**
+ * Kick the radio's watchdog
+ * @note This function might not be implemented for all radios
+ * @return KRadioStatus RADIO_OK if OK, error otherwise
+ */
+KRadioStatus k_radio_watchdog_kick(void);
```
jacoffey3: This is probably not transferable at all. I'd leave it in the ISIS HAL, or make it a type of reset.

docs/device-api/radio.rst
```diff
@@ -0,0 +1,118 @@
+Using a Radio
```
jacoffey3: This, honestly, is going to change a lot as we integrate more radios. I wouldn't worry too much about trying to account for new radios, because as we integrate them we'll just need to revise it.

radio/radio-api/radio-api/radio.h
```diff
@@ -0,0 +1,117 @@
...
+{
+    /**
+     * Six character station call-sign
+     */
+    uint8_t ascii[6];
+    /**
+     * One byte station SSID value
+     */
+    uint8_t ssid;
+} ax25_callsign;
```
plauche: I'm not sure about having the `ax25_callsign` here. We probably want to either pushing framing information down into the radio specific hals (in this case the isis radio forces ax25 and we don't see it in the flight software) or handle it at a higher controller level.

radio/radio-api/radio-api/radio.h
```diff
@@ -0,0 +1,117 @@
...
+
+/**
+ * Radio reset types
+ */
+typedef enum {
+    /** Perform hardware-level radio reset */
+    RADIO_HARD_RESET,
+    /** Perform software radio reset */
+    RADIO_SOFT_RESET
+} radio_reset;
```
plauche: This enum's type doesn't quite match our other types. Maybe it should be `KRadioReset`?

docs/device-api/radio.rst
CatGarab: :+1: That's totally fine. The documentation just needs to start somewhere

radio/radio-api/radio-api/radio.h
CatGarab: I'd like to leave it here for now, since the AX.25 call-sign is a) specifically a radio construct and b) a standard.

radio/radio-api/module.json
```diff
@@ -0,0 +1,10 @@
+{
+    "license":"Apache-2.0",
+    "name":"radio-api",
+    "repository":{
+        "url":"git://github.com/kubostech/radio-api",
+        "type":"git"
+    },
+    "version":"0.1.0",
```
plauche: This config file should probably have the dependency for the `trxvu-radio-api` so that this top level api can be used without manual modifications.

radio/radio-api/radio-api/radio.h
```diff
@@ -0,0 +1,111 @@
...
+ * @return KRadioStatus RADIO_OK if OK, error otherwise
+ */
+KRadioStatus k_radio_reset(uint8_t type);
+/**
+ * Send a message to the radio's transmit buffer
+ * @param [in] buffer Pointer to the message to send
+ * @param [in] len Length of the message to send
+ * @return uint8_t See specific radio API documentation for return code documentation
+ */
+uint8_t k_radio_send(char * buffer, int len);
```
plauche: Ok don't hate me but now that I'm looking at the actual implementation of this interface I wonder if this should also return a `KRadioStatus` and take another pointer parameter for the specific send return code?

CatGarab: Yeah, I'd been kind of thinking that, too. I'll add it

PR #196, opened 2017-10-03
Documenting windows powershell requirement

CatGarab:
As of v1.9.5, Vagrant now requires Windows operating systems to have at least Windows PowerShell v3.0. By default, Windows 7 SP1 comes with v2.0. I updated the docs to list the requirement and the steps necessary to upgrade PowerShell.

Sadly Windows Vista is no longer a supported OS, since it only supports up to PowerShell v2.0...

### DIFF
--- a/docs/sdk-installing.rst
+++ b/docs/sdk-installing.rst
@@ -73,17 +73,7 @@ and running the following commands::
 
 The ``Major`` field should have a value of atleast ``3``.
 
-Currently, the latest stable release of Windows PowerShell is available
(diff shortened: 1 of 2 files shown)


docs/sdk-installing.rst
```diff
@@ -40,13 +40,51 @@ System Requirements
...
+    Major  Minor  Build  Revision
+    -----  -----  -----  --------
+    2      0      -1     -1
+  
+    
+    PS $> exit
+
+The ``Major`` field should have a value of atleast ``3``.
+
+Currently, the latest stable release of Windows PowerShell is available
```
jacoffey3: Do we want to document how to update powershell? Is there not a good external resource for that?

PR #197, opened 2017-10-05
Minor doc tweaks

CatGarab:
1. BBB buildroot config name is wrong
2. If currently running KubOS Linux, `umount` command must be run twice in order to successfully transfer the new SD image to the eMMC

### DIFF
--- a/docs/kubos-linux-on-bbb.rst
+++ b/docs/kubos-linux-on-bbb.rst
@@ -337,6 +337,7 @@ To flash the eMMC, log into the board and then run these commands:
 ::
 
     $ umount /dev/mmcblk*
+    $ umount /dev/mmcblk*
     $ dd if=/dev/mmcblk0 of=/dev/mmcblk1
     
 The four status LEDs on the board should start flashing in a random pattern. This indicates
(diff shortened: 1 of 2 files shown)


CatGarab:
@jacoffey3 The two umounts are due to the way the directories are mapped.
`/home` maps to mmcblk1p3.
`/home/microsd` maps to mmcblk0p2, but since it's technically a subdirectory of `/home`, the first umount will only unmount mmcblk0p2.
So umount 1 does everything except mmcblk1p3, umount 2 finishes it

jacoffey3:
@CatGarab Ahh I get it. Might put a note in there for why both are needed.

docs/kubos-linux-on-bbb.rst
```diff
@@ -337,8 +337,14 @@ To flash the eMMC, log into the board and then run these commands:
 ::
 
     $ umount /dev/mmcblk*
+    $ umount /dev/mmcblk*
     $ dd if=/dev/mmcblk0 of=/dev/mmcblk1
     
+.. note::
```
plauche: I think it would be a bit more clear if we made the two unmount commands explicit to either the devices or the folders. Something like...
`$ umount /home/microsd`
`$ umount /home`
I understand the need to do both, but it is confusing seeing two identical `umount` calls which both attempt to unmount all mmcblk devices.

PR #200, opened 2017-10-09
Doc Re-Org

CatGarab:
**Problem**
There are too many links on the left-hand menu of our docs page (I shouldn't have to scroll to find the link that I want). This is irritating for me and potentially overwhelming for new users.
**Proposed Solution** 
A less-flat doc organization

The generated version of the docs can be found at http://docs.kubos.co/sphinx/

TODO: Integrate the evented architecture stuff that will be merged into master soon


CatGarab:
Potentential change - Break out the API docs into two different sections: the 'How-to' guides, and the actual APIs

marshall:
wow, PR number 200. 🎉  congrats team!

docs/apis.rst
```diff
@@ -0,0 +1,22 @@
+Kubos APIs
+==========
+
+Kubos Core: APIs for supported sensors
+
+Kubos HAL: Hardware interface abstractions (I2C, SPI, etc)
+
+Kubos IPC: w/e this is
+
+Kubos Telemetry: telemetry stuff...
```
marshall: most of these look like placeholders.. finish them up before pushing?

docs/faq-troubleshooting.rst
```diff
@@ -9,7 +9,7 @@ FAQs
 How do I contact y'all?
 ~~~~~~~~~~~~~~~~~~~~~~~
 
-Our community Slack: slack.kubos.co
+Our community Slack: `slack.kubos.co <http://slack.kubos.co>`__
```
marshall: should be https://slack.kubos.co I think

docs/schema.json
```diff
@@ -0,0 +1,117 @@
+{
```
marshall: I like this! would be cool if we moved to an actual JSON schema in the future too http://json-schema.org/examples.html

jessehamner: I agree -- JSON is useful stuff, and human readable.

docs/faq-troubleshooting.rst
davesims: @marshall @CatGarab Yep. Without it, port 80 redirects to http://docs.kubos.co/1.0.1/index.html

docs/sdk-project-config.rst
```diff
@@ -201,7 +201,7 @@ If present, the ``fs`` file system structure enables support for accessing stora
     using FatFS over SDIO if it is available on your target.`
     
     :property dev: SPI bus the device is connected to
-    :proptype dev: :cpp:enum:`KSPINum`
+    :proptype dev: :cpp:type:`KSPINum`
```
davesims: Out of curiosity, what's the significance of the shift from `enum` to `type`?

CatGarab: Sphinx complaining that the variable is actually a type because it's a `typedef enum` rather than just `enum`. So I changed it to get rid of the warning :stuck_out_tongue:

docs/schema.json
CatGarab: whoops, that file wasn't supposed to make it into this. It was a local file I was using for reference. But yes, I would love to have a schema for our project configuration options.

CatGarab:
Everything is polished and ready for actual review

jacoffey3:
Comments from Mitch: 
image

CatGarab:
@jacoffey3 @mitch929 We use the [Read the Docs' theme](https://github.com/rtfd/sphinx_rtd_theme), which controls most of the things those comments are about. It can be modified, but requires knowledge of CSS (and maybe Sass), which I don't have. So the changes are probably small, but I have no idea how long it would actually take.

Info about themes: http://www.sphinx-doc.org/en/stable/theming.html

jacoffey3:
Ahh gotcha. I think we're okay for now, but we'll keep the changes on file for later.

docs/index.rst
```diff
@@ -1,122 +1,64 @@
...
 libraries necessary to quickly bring up space ready software. We
 leverage multiple existing open source projects like FreeRTOS and CSP,
 along with our own custom framework and SDK.
 
-Looking to build an application on Kubos? Check out our :doc:`KubOS RT 
-quick start guide <first-rt-project>` or our :doc:`KubOS Linux quick 
-start guide <first-linux-project>`
+Looking to build an application on Kubos? Check out our quick start guides:
+
+ - :doc:`KubOS RT quick start guide <rt-docs/first-rt-project>`
```
plauche: nit: Linux should probably come before RT

docs/rt-docs/first-rt-project.rst
```diff
@@ -7,7 +7,7 @@ using the Kubos SDK.
 Prerequisites
```
plauche: I can't comment on the line but we link to an example Linux project on this RT-specific page.

docs/linux-docs/first-linux-project.rst
```diff
@@ -7,7 +7,7 @@ using the Kubos SDK.
 Prerequisites
```
plauche: Can we add a list of the available sample Linux projects to this page?

docs/index.rst
```diff
@@ -1,122 +1,64 @@
-Kubos
-~~~~~
+
+.. figure:: images/kubos_logo.png
+    :align: center
+
+Overview
+========
```
plauche: I wonder if we should link to the changelog/version info on this page.

PR #204, opened 2017-10-30
Radio API - Moving away from generic pointers

CatGarab:
Passing generic pointers is bad. Passing pointers to context-specific structures is better.

Updating the generic radio API to include the appropriate specific radio API based on the `config.json` file so that quasi-generic radio structures can be used instead of generic pointers.

The TRXVU API will also be updated to reflect these changes.

### DIFF
--- a/radio/radio-api/radio-api/radio.h
+++ b/radio/radio-api/radio-api/radio.h
@@ -25,17 +25,6 @@
 #include "radio-struct.h"
 #include "radio-impl.h"
 
-/* Define the radio-specific structures/enumerators */
-/*
-#if defined(YOTTA_CFG_RADIO_TRXVU) && !defined(HAVE_RADIO)
-#define HAVE_RADIO
-#include <trxvu-radio-api/radio.h>
-#endif
-#ifndef HAVE_RADIO
-#error No radio defined!
-#endif
-*/
-
 /* Define the global functions */
 /**
  * Initialize the radio interface
--- a/radio/radio-api/radio-api/radio.h
+++ b/radio/radio-api/radio-api/radio.h
@@ -22,51 +22,15 @@
 #pragma once
 
 #include <stdint.h>
-
-/* Define the global structures/enumerators */
-/**
- * Radio function return values
- */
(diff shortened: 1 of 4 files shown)


CatGarab:
[TRXVU Telemetry Calculations](https://trello.com/c/KPI8h8VS/35-trxvu-telemetry-calculations)

radio/radio-api/radio-api/radio.h
```diff
@@ -22,46 +22,18 @@
...
-    uint8_t ascii[6];
-    /**
-     * One byte station SSID value
-     */
-    uint8_t ssid;
-} ax25_callsign;
+/* Define the radio-specific structures/enumerators */
+#if defined(YOTTA_CFG_RADIO_TRXVU) && !defined(HAVE_RADIO)
+#define HAVE_RADIO
+#include <trxvu-radio-api/radio.h>
```
plauche: I believe that we could just name the header in the implementation module something else, and as long as we keep the name the same in all of the radio implementations we can just use one include. So if we called it `radio_impl.h`, then I think we could just do

```
#include "radio_impl.h"
```
once and it would work for all dependencies which have that file. We do the same thing in the gpio part of `kubos-hal` with the `pins.h` header.

CatGarab: :tada: That's way more elegant

radio/radio-api/radio-api/radio.h
```diff
@@ -22,46 +22,21 @@
...
-    /**
-     * One byte station SSID value
-     */
-    uint8_t ssid;
-} ax25_callsign;
+/* Define the radio-specific structures/enumerators */
+/*
+#if defined(YOTTA_CFG_RADIO_TRXVU) && !defined(HAVE_RADIO)
+#define HAVE_RADIO
+#include <trxvu-radio-api/radio.h>
```
plauche: nit: commented out code

CatGarab: :laughing: shame on me

PR #211, opened 2017-11-15
Example python-based subsystem handler

plauche:
- [x] Create example python based handler
  - [x] Present GraphQL + GraphiQL endpoints
  - [x] Example subsystem class
  - [x] Handle query accessing subsystem


plauche:
[Create generic example python handler (graphql query)](https://trello.com/c/tHP3FNK3/84-create-generic-example-python-handler-graphql-query)

PR #213, opened 2017-11-16
Added mutations into example.

plauche:
- [x] Expose mutation over GraphQL
- [x] Modify model to handle mutation functions


plauche:
[Add graphql mutations to python handler example](https://trello.com/c/MWYDaeJV/85-add-graphql-mutations-to-python-handler-example)

plauche:
@jacoffey3 for this mutation implementation I took a slightly different approach. Mutations here return back a `Status` model which should indicate the pass/fail/return code of the mutation function, they also have the ability to query the state of the `Subsystem` model inside of the `Status`. Thoughts on that approach vs just returning back the `Subsystem` model?

examples/python-handler/handler/models.py
```diff
@@ -26,3 +26,23 @@ def refresh(self):
 
         print "Querying for subsystem status"
         self.power_on = not self.power_on
+
+    def set_power_on(self, power_on):
+        """
+        Controls the power state of the subsystem
+        """
+
+        print "Sending new power state to subsystem"
```
CatGarab: nit: Since this is an example program, it might be nice to print the old and new values in this message

examples/python-handler/handler/models.py
```diff
@@ -26,3 +26,23 @@ def refresh(self):
...
+
+
+class Status(graphene.ObjectType):
+    """
+    Model representing execution status. This allows us to return
+    the status of the mutation function alongside the state of
+    the model affected.
+    """
+
+    status = graphene.Boolean()
```
CatGarab: Just making sure I'm looking at this example the right way: this field would potentially be the return code from our underlying handler/API code, right? (For example, if there was an I2C error communicating with whatever subsystem it would bubble up to here)

examples/python-handler/handler/schema.py
```diff
@@ -31,4 +31,34 @@ def resolve_subsystem(self, info):
...
+
+    Output = Status
+
+    def mutate(self, info, power_on):
+        """
+        Handles request for subsystem powerOn mutation
+        """
+
+        status = Status(status=True, subsystem=_subsystem)
+        if power_on != None:
```
CatGarab: This should probably throw an error

plauche: Do you mean it should throw an error if `power_on` is `None`? That is actually a correct state, as it means the graphql mutation does not affect `power_on`.

examples/python-handler/handler/models.py
plauche: Yup that is the idea. In this case it doesn't make a lot of sense for the mutation to just return back the state of the subsystem without any indication of the return/status of the function called.

examples/python-handler/handler/schema.py
plauche: Yup that is correct.

PR #215, opened 2017-11-29
Added info and graphics about potentially concerning messages.

jessehamner:
Added two screencaps and some text to explain them, based on some system output during BBB Linux installation.


docs/installation-docs/installing-linux-bbb.rst
```diff
@@ -135,6 +135,15 @@ To flash the eMMC, log into the board and then run these commands:
     $ umount /home
     $ dd if=/dev/mmcblk0 of=/dev/mmcblk1
     
+.. figure:: ../images/kubos_bbb_linux_dd.png
+   :alt: dd complaints.
+
+It is possible that you will see some errors when you try to unmount the directories. 
+That's likely not a problem. The ``dd`` process will hang for a long time at the "nonblocking pool" line.
```
CatGarab: The `random: nonblocking pool is initialized` is actually an informational message that's part of Linux initialization. It's issued 10 seconds or so after the rest of the OS comes up. So you can add it to the list of "random messages that you might see that are totally inconsequential", but it's not actually associated with the `dd` command.

docs/installation-docs/installing-linux-bbb.rst
```diff
@@ -135,6 +135,15 @@ To flash the eMMC, log into the board and then run these commands:
     $ umount /home
     $ dd if=/dev/mmcblk0 of=/dev/mmcblk1
     
+.. figure:: ../images/kubos_bbb_linux_dd.png
+   :alt: dd complaints.
+
+It is possible that you will see some errors when you try to unmount the directories. 
+That's likely not a problem. The ``dd`` process will hang for a long time at the "nonblocking pool" line. 
+Then, the system will complain that there is no space left on the device. 
+You can safely ignore this error if the SD card is 4GB or larger. These errors will
```
CatGarab: The SD card should always be atleast 4GB, since that's the required size of the KubOS Linux image file. It's more accurate to say something like "While the eMMC is 4GB, a small portion is set up as read-only and dedicated to boot-time processing. Since the SD card will be larger than the writeable area of the eMMC, the 'No space left on device' message will be issued."

docs/installation-docs/installing-linux-bbb.rst
```diff
@@ -152,16 +161,26 @@ Re-Flash the SD Card
...
 
     If you do not have a microSD card in the board, the system will not boot.
 
+
+.. figure:: ../images/kubos_bbb_linux_mount_errors.png
+   :alt: mount complaints during boot.
+
+The boot messages may cause concern, in that it looks like they indicate that
+the system is having trouble mounting directories. You can most likely ignore
+these warnings as well.
```
CatGarab: Be specific about which "errors" are actually okay. Maybe something like this:
"You will see messages as the data partitions are mounted. For example, `EXT4-fs (mmcblk0p2): couldn't mount as ext3 due to feature incompatibilities`. While they may seem like errors, they are a normal part of the boot process as the system detects the partition file type. If there are actual issues mounting a partition, the resulting error message will look like this: `mount: mounting /dev/mmcblk1p3 on /home failed: No such file or directory`

docs/installation-docs/installing-linux-bbb.rst
jessehamner: Done.

docs/installation-docs/installing-linux-bbb.rst
jessehamner: Done.

docs/installation-docs/installing-linux-bbb.rst
jessehamner: Have addressed this as well.

jessehamner:
Have untracked all the files requested. Ready to go.

PR #216, opened 2017-11-29
Generic ADCS API

CatGarab:
**PR Overview**:

Creating a generic ADCS API

**Documentation**:

All documentation has been included in this PR

**Testing**:
Not needed since this is only a header file


CatGarab:
[ADCS API](https://trello.com/c/NANUoXIv/112-adcs-api)

adcs/adcs-api/adcs-api/adcs.h
```diff
@@ -0,0 +1,119 @@
...
+/**
+ * Terminate the ADCS interface
+ */
+void k_adcs_terminate(void);
+/**
+ * Execute ADCS no-op command
+ * @note This function might not be implemented for all ADCSs.
+ * @return KADCSStatus ADCS_OK if OK, error otherwise
+ */
+KADCSStatus k_adcs_noop();
```
plauche: nit: the parameter here can be `void`

docs/apis/device-api/adcs.rst
```diff
@@ -0,0 +1,184 @@
...
+    
+    // Project logic
+    
+    k_adcs_terminate();
+
+Run-Time ADCS Configuration
+---------------------------
+
+Some ADCSs may support or require that the ADCS be configured during run-time. This can be done by passing a pointer to 
+a JSON structure (abstracted by ``adcs_config``) to the ``k_adcs_configure`` function.
```
plauche: Where is `adcs_config` defined? I would expect it in one of the `adcs-api` header files here but it doesn't seem to be there.

plauche: Ok found it in the imtq specific library, I think it would be more appropriately defined here at the top level, assuming we want all implementations of adcs configuration and telemetry to use json.

CatGarab: That is an excellent question.

@plauche @jacoffey3  Do we want all implementations of ADCS config/telemetry to use JSON?

jacoffey3: Yes we do, but the specific configuration parameters are going to change unit to unit.

PR #217, opened 2017-12-01
Example generic rust-based subsystem handler

plauche:
- [x] Create example rust based handler
  - [x] Present GraphQL + GraphiQL endpoints
  - [x] Example subsystem model
  - [x] Handle query accessing subsystem
  - [x] Example functions for fetching subsystem data

### DIFF
--- a/examples/rust-handler/src/model.rs
+++ b/examples/rust-handler/src/model.rs
@@ -14,10 +14,22 @@
 // limitations under the License.
 //
 
-use juniper::FieldResult;
+use std::fmt;
+use std::error;
 
 /// Model for handler's subsystem
 pub struct Subsystem;
+
+#[derive(Debug)]
+pub enum SubsystemError {
+    IOFailure,
+}
+
+impl fmt::Display for SubsystemError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "{}", self)
+    }
+}
 
 impl Subsystem {
     /// Creates new Subsystem structure instance
@@ -31,16 +43,18 @@ impl Subsystem {
     /// Power status getter
     /// Code querying for new power value
     /// could be placed here
-    pub fn power(&self) -> FieldResult<bool> {
+    pub fn power(&self) -> Result<bool, SubsystemError> {
         println!("getting power");
         // Low level query here
         Ok(true)
+        // If we had a failure retrieving power status...
+        // Err(SubsystemError::IOFailure)
     }
 
     /// Uptime getter
     /// Code querying for new uptime value
     /// could be placed here
-    pub fn uptime(&self) -> FieldResult<i32> {
+    pub fn uptime(&self) -> Result<i32, SubsystemError> {
         println!("getting uptime");
         // Low level query here
         Ok(100)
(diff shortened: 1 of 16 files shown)


plauche:
[Create generic example rust handler (graphql query)](https://trello.com/c/4QDkVx70/89-create-generic-example-rust-handler-graphql-query)

examples/rust-handler/src/schema.rs
```diff
@@ -0,0 +1,57 @@
...
+impl Context {
+    /// Give us a reference to subsystem for passing
+    /// along the Juniper chain
+    pub fn get_subsystem(&self) -> &Subsystem {
+        &self.subsystem
+    }
+}
+
+/// GraphQL model for Subsystem
+graphql_object!(Subsystem: Context as "Subsystem" |&self| {
```
CatGarab: Could you roll the `impl Subsystem` work into this?

examples/rust-handler/src/schema.rs
```diff
@@ -0,0 +1,57 @@
...
+    field uptime() -> i32 as "Uptime of subsystem" {
+        self.uptime()
+    }
+});
+
+pub struct QueryRoot;
+
+/// GraphQL model for base query
+graphql_object!(QueryRoot : Context as "Query" |&self| {
+    field subsystem(&executor) -> Option<&Subsystem>
```
CatGarab: This should likely return a `FieldResult` instead. For a real implementation, it's possible that queries will fail for some reason.

examples/rust-handler/src/model.rs
```diff
@@ -0,0 +1,50 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+/// Model for handler's subsystem
+pub struct Subsystem {
+    power: bool,
```
CatGarab: You could use the `#[graphql(description="..."]` statements here to define what the fields are rather than "as [blah]" statements later. This would be similar to how we document structures in our C code.

jacoffey3:
This is like Greek to me. I might want your help deciphering this at some point Ryan, otherwise it'll be useless for me to review.

examples/rust-handler/src/schema.rs
plauche: I'm not totally sure what you mean. We are using the `Subsystem` struct here which already has the functions provided by `impl Subsystem`. The `graphql_object` macro is just implementing a new trait and set of fields on the struct alongside those already created. Most of the Juniper examples I've seen are structured this way, thin method calls in the `graphql_object` fields and more logic in the`models` file.

plauche:
@CatGarab currently you can query for power and uptime independently. The `power()` or `uptime()` functions will only be called when that attribute is queried.

examples/rust-handler/src/model.rs
plauche: I looked over the (new :smile:) documentation for using those `#[graphql` and `#[derive(GraphQLObject)]` macros and it looks like they only work for structs with fields which are either primitives or other GraphQL types. I don't know if we can use them to bring in struct function fields, which is primarily what we'll be using to retrieve information from the subsystem. I could easily see using a struct without any primitive fields and just functions returning back live data.

examples/rust-handler/src/schema.rs
plauche: :+1:

PR #218, opened 2017-12-04
Add mutations into example rust handler

plauche:
- [x] Expose mutation over GraphQL
- [x] Modify model or add functions to handle mutation
- [x] Demonstrate returning back actual error conditions from mutations


plauche:
[Add graphql mutations to rust handler example.](https://trello.com/c/6j134ur1/90-add-graphql-mutations-to-rust-handler-example)

examples/rust-handler/src/schema.rs
```diff
@@ -45,13 +45,43 @@ graphql_object!(Subsystem: Context as "Subsystem" |&self| {
...
+
+/// Base GraphQL mutation model
+graphql_object!(MutationRoot : Context as "Mutation" |&self| {
+
+    // Each field represents functionality available
+    // through the GraphQL mutations
+    field power(&executor, state: bool) -> Option<Status>
+        as "Set power state of subsystem"
+    {
+        Some(executor.context().get_subsystem().set_power(state))
```
CatGarab: This, and the parent function, should return a `FieldResult` (or just a `Result`?) to pass back any underlying errors.

examples/rust-handler/src/model.rs
```diff
@@ -40,11 +52,29 @@ impl Subsystem {
         self.power
     }
 
+    /// Power state setter
+    /// Here we would call into the low level
+    /// device function
+    pub fn set_power(&self, _power: bool) -> Status {
+        println!("Setting power state");
+        // Send command to device here
+        Status { status: true }
```
CatGarab: Why not manipulate the actual `power` field, since that's what this function's for?

examples/rust-handler/src/model.rs
```diff
@@ -14,6 +14,18 @@
 // limitations under the License.
 //
 
+
+/// Model for subsystem function status
+pub struct Status {
```
CatGarab: Why isn't this just a field of `Subsystem`?

plauche:
@CatGarab I went back and forth on the granularity of the mutations, glad to have another opinion :smile:. 

I created some more abstract return types for the mutations similar to what the Apollo blog suggested. They make things a little verbose, but I can see how they could provide more flexibility in the future and it does seem like GraphQLy.

PR #219, opened 2017-12-06
Rust handler calling C library

plauche:
- [x] Create dummy C library for demonstration
- [x] Create basic Rust binding to dummy C library
- [x] Call binding from inside of `Subsystem` model functions
- [x] Create crate for easy C library integration


plauche:
[Compile rust example handler against c/yotta/kubos module](https://trello.com/c/v4ce4fZX/119-compile-rust-example-handler-against-c-yotta-kubos-module)

examples/rust-handler/Cargo.toml
```diff
@@ -9,3 +9,5 @@ juniper_iron = "0.1.1"
 iron = "0.5.1"
 mount = "0.3.0"
 logger = "0.3.0"
+
+kubos-hal-iobc = { path = "../../hal/kubos-hal-iobc" }
```
CatGarab: super nit: I now dislike "rust-generic-handler" as the name for this. You're adding calls for the iOBC supervisor which is highly specific, rather than "generic".

targets/target-x86-linux-native/CMake/toolchain.cmake
```diff
@@ -4,8 +4,8 @@ cmake_minimum_required(VERSION 2.8)
 
 # default to C99
 # Enable extra warnings for debug builds
-set(CMAKE_C_FLAGS_DEBUG "-std=gnu99 -Wall -Wextra -Wshadow  -Wstrict-overflow -fno-strict-aliasing -Wno-unused-parameter" CACHE STRING "")
-set(CMAKE_C_FLAGS "-std=gnu99" CACHE STRING "")
+set(CMAKE_C_FLAGS_DEBUG "-fPIC -std=gnu99 -Wall -Wextra -Wshadow  -Wstrict-overflow -fno-strict-aliasing -Wno-unused-parameter" CACHE STRING "")
+set(CMAKE_C_FLAGS "-fPIC -std=gnu99" CACHE STRING "")
```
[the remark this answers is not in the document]
plauche: This flag causes gcc to build position-independent code. Adding it was necessary to allow Rust to link against libraries produced by gcc. The rust compiler (rustc) by default builds binaries and libraries which are position independent. The flag will also make it possible in the future to link (yotta built) libraries into shared libraries, whether built in Rust or C.

I believe this flag wasn't originally included in the Kubos/Yotta toolchain due to the embedded origins of Yotta. It wouldn't really make sense to build a position independent library (or binary) in the embedded/RTOS context, as you are mostly dealing with fixed static memory locations.

examples/rust-handler/Cargo.toml
plauche: Yup agreed. I'll try to find something more generic to call...

hal/kubos-hal-iobc/build.rs
```diff
@@ -0,0 +1,38 @@
+/// A custom build.rs for building out the kubos-hal-iobc
```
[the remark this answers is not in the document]
plauche: :+1:

plauche:
[Cross-compile and run example rust handler on iobc](https://trello.com/c/cXXRaGyi/93-cross-compile-and-run-example-rust-handler-on-iobc)

targets/target-x86-linux-native/CMake/toolchain.cmake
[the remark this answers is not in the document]
CatGarab: :+1:

kubos-build-helpr/Cargo.toml
```diff
@@ -0,0 +1,6 @@
+[package]
+name = "kubos-build-helpr"
```
CatGarab: nit: It's bugging me that you dropped the 'e'. I'd love it if this was "kubos-build-helper" instead

hal/kubos-hal-iobc/Cargo.toml
```diff
@@ -0,0 +1,8 @@
+[package]
+name = "kubos-hal-iobc"
```
CatGarab: It looks like `extern-lib` replaced this module as the example Kubos library. If that's the case, and `kubos-hal-iobc` is no longer being linked in, then these Rust files can be removed.

kubos-build-helpr/src/lib.rs
```diff
@@ -0,0 +1,86 @@
...
+/// extension which takes a kubos target and then
+/// determines the correct Rust triplet
+fn determine_target() -> String {
+    match env::var("TARGET").unwrap().as_ref() {
+        // Default native vagrant target
+        "x86_64-unknown-linux-gnu" => String::from("x86-linux-native"),
+        // Currently set to the beaglebone black toolchain
+        // We will eventually need to determine if we intend
+        // to build with the pumpkin-mbm2 toolchain
+        "arm-unknown-linux-gnueabidf" => String::from("kubos-linux-beaglebone-gcc"),
```
CatGarab: FYI: the BBB and MBM2 have the exact same toolchain (and, theoretically, always will), since the MBM2 basically just builds on top of the BBB.
The two different Kubos SDK targets are really just a formality to prevent users from getting confused about why there isn't explicitly a MBM2 target.

kubos-build-helpr/src/lib.rs
```diff
@@ -0,0 +1,86 @@
...
+    }
+}
+
+/// Performs the actual `kubos build` process
+///
+/// We also do a `kubos clean` beforehand because we
+/// aren't building in a way that Cargo is aware of
+/// so it isn't able to clean up for us
+fn do_build(target: &str) {
+    // Can we attach this to cargo clean?
```
CatGarab: Need a `kubos link -a` step

PR #222, opened 2017-12-20
Adding cargo `kubos build` subcommand

plauche:
- [x] Create cargo sub command to help handle "kubos build" related work
  - [x] Translates kubos target into rust/clang triplet
  - [x] Passes kubos target along for consumption by build scripts

Once this PR is merged into master the `cargo kubos` command can be installed into the Vagrant box.


cargo-kubos/src/main.rs
```diff
@@ -0,0 +1,87 @@
...
+    let args: Vec<String> = env::args().collect();
+
+    let mut opts = Options::new();
+    opts.optflag("b", "build", "builds module");
+    opts.optopt("t", "target", "sets (kubos) build target", "NAME");
+    let matches = match opts.parse(&args[1..]) {
+        Ok(m) => m,
+        Err(f) => panic!(f.to_string()),
+    };
+    if matches.opt_present("b") {
```
CatGarab: If `-b` is always required, it doesn't need to be an command option

cargo-kubos/src/main.rs
```diff
@@ -0,0 +1,87 @@
...
+        exit(status.code().unwrap());
+    }
+}
+
+fn main() {
+    let args: Vec<String> = env::args().collect();
+
+    let mut opts = Options::new();
+    opts.optflag("b", "build", "builds module");
+    opts.optopt("t", "target", "sets (kubos) build target", "NAME");
```
CatGarab: nit: capitalize Kubos

cargo-kubos/src/main.rs
```diff
@@ -0,0 +1,87 @@
...
+            )
+        }
+    }
+}
+
+/// Perform `cargo build` using the proper
+/// Rust/Clang target triplet
+fn cargo_build(target: &str) {
+    let status = Command::new("cargo")
+        .args(&["build", "--target", target, "-vv"])
```
CatGarab: It would be nice if you could pass through any `-v` parameters, instead of having it on by default

cargo-kubos/src/main.rs
marshall: yeah, it looks like `-b` is completely extraneous :)

cargo-kubos/src/main.rs
```diff
@@ -0,0 +1,87 @@
...
+        let k_target = match matches.opt_str("t") {
+            Some(t) => t,
+            None => String::from("x86-linux-native"),
+        };
+        let c_target = target_converter(&k_target);
+        env::set_var("CARGO_KUBOS_TARGET", &k_target);
+        cargo_build(&c_target);
+    } else {
+        println!("cargo-kubos is a helper utility for building");
+        println!("kubos/yotta based modules under Cargo. It is");
```
marshall: nit again: Kubos

cargo-kubos/src/main.rs
```diff
@@ -0,0 +1,87 @@
...
+        println!("primarily used when building crates which either");
+        println!("contain a yotta module or depend on one.");
+        println!("");
```
marshall: Should everything from here below be generated by [Options::usage()](https://doc.rust-lang.org/getopts/getopts/struct.Options.html#method.usage)?

cargo-kubos/src/main.rs
```diff
@@ -0,0 +1,87 @@
...
+    opts.optflag("b", "build", "builds module");
+    opts.optopt("t", "target", "sets (kubos) build target", "NAME");
+    let matches = match opts.parse(&args[1..]) {
+        Ok(m) => m,
+        Err(f) => panic!(f.to_string()),
+    };
+    if matches.opt_present("b") {
+        let k_target = match matches.opt_str("t") {
+            Some(t) => t,
+            None => String::from("x86-linux-native"),
```
marshall: nit: since this is the same key used in the `target_converter` matcher, can you change `"x86-linux-native"` in both places to a single `&'static str` value?

kubos-build-helpr/src/lib.rs
```diff
@@ -1,34 +1,38 @@
...
 
-/// Translates Cargo/Rust/Clang target triplet
-/// into the proper target for `kubos build`
-///
-/// This bit might be better wrapped in a cargo
-/// extension which takes a kubos target and then
-/// determines the correct Rust triplet
+/// Retrieve the kubos target from the environment
+/// variable CARGO_KUBOS_TARGET. This variable *should*
+/// get set when the `cargo kubos` command is run.
```
marshall: good comments 👍

cargo-kubos/src/main.rs
```diff
@@ -0,0 +1,87 @@
...
+        .stdout(Stdio::inherit())
+        .stderr(Stdio::inherit())
+        .status()
+        .unwrap();
+
+    // Attempt to exit in a way which
+    // honors the subprocess exit code
+    if status.success() {
+        exit(1)
+    } else {
```
marshall: nit: the surrounding `else` here is unnecessary

cargo-kubos/src/main.rs
```diff
@@ -0,0 +1,87 @@
...
+fn target_converter(kubos_target: &str) -> String {
+    match kubos_target {
+        "x86-linux-native" => String::from("x86_64-unknown-linux-gnu"),
+        "kubos-linux-beaglebone-gcc" => String::from("arm-unknown-linux-gnueabihf"),
+        "kubos-linux-pumpkin-mbm2-gcc" => String::from("arm-unknown-linux-gnueabihf"),
+        "kubos-linux-isis-gcc" => String::from("arm-unknown-linux-gnueabi"),
+        _ => {
+            panic!(
+                "Target not supported for cargo/yotta builds {}",
+                kubos_target
```
marshall: Does it make sense to add a list of targets that are supported to this error message?

plauche:
@CatGarab re: putting this in `kubos-cli`, I agree that it may belong in a different repo, but I have mixed feelings about it landing in `kubos-cli`.  That tool is very much centered around `yotta` and extending it`s functionality. I see this PR and `cargo` as a beginning of moving away from our `yotta` based tooling to something more appropriate for our Linux development. Thoughts @kyleparrott?

kyleparrott:
I think this is an ok place for now. I don't think this belongs in the Kubos-cli repo.

In time I'd like to see this move into the direction of becoming a new repo/cli project that will eventually supersede the current kubos-cli.

plauche:
I've changed the functionality of this subcommand to make it more of a generic pass-through to Cargo. It no longer builds by default, but will take any cargo sub command via `-c command`.

cargo-kubos/src/main.rs
```diff
@@ -0,0 +1,117 @@
...
+    } else {
+        Vec::new()
+    };
+
+    if matches.opt_present("h") {
+        print_usage(opts);
+    } else {
+        let k_target = match matches.opt_str("t") {
+            Some(t) => t,
+            None => String::from("x86-linux-native"),
```
marshall: nit: this string literal is still used in 2 places (above in `target_converter` and here), can you make the value const and use it in both places?

cargo-kubos/src/main.rs
```diff
@@ -0,0 +1,117 @@
...
+
+extern crate getopts;
+
+use getopts::Options;
+use std::process::{exit, Command, Stdio};
+use std::env;
+
+/// Take a kubos target and convert it
+/// to a Rust/Clang target triplet
+fn target_converter(kubos_target: &str) -> String {
```
marshall: much better 👍

PR #224, opened 2018-01-03
Adding null checking to user-facing JSON functions

CatGarab:
The JSON library's user-facing functions don't currently do any argument checking, which can lead to segfaults. This PR fixes that.

The user-facing functions are all the ones that start with `json_`, as seen in `ccan/json/json/json.h`

### DIFF
--- a/ccan/json/source/json.c
+++ b/ccan/json/source/json.c
@@ -365,7 +365,7 @@ JsonNode *json_decode(const char *json)
     }
 
 	const char *s = json;
-	JsonNode *ret;
+	JsonNode *ret = NULL;
 	
 	skip_space(&s);
 	if (!parse_value(&s, &ret)) {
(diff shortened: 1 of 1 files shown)


ccan/json/source/json.c
```diff
@@ -360,12 +360,18 @@ static bool number_is_valid(const char *num);
 
 JsonNode *json_decode(const char *json)
 {
+    if (json == NULL) {
```
plauche: nit: spacing isn't consistent

ccan/json/source/json.c
```diff
@@ -360,12 +360,18 @@ static bool number_is_valid(const char *num);
...
+        return NULL;
+    }
+
 	const char *s = json;
 	JsonNode *ret;
 	
 	skip_space(&s);
-	if (!parse_value(&s, &ret))
+	if (!parse_value(&s, &ret)) {
+	    json_delete(ret);
```
plauche: It looks like in the cases when `parse_value` returns `false` it has not set `ret` to any value. Calling `json_delete` here could cause issues since `ret` hasn't been initialized to `NULL` and had no value set inside of `parse_value`.

CatGarab: Good catch!

PR #227, opened 2018-01-12
[WIP] iOBC Supervisor Service

plauche:
This work exposes the supervisor API from `kubos-hal-iobc` over a GraphQL interface.

- [ ] Change name/folder "handler" -> "service"
- [ ] Investigate better way to exposing test data

### DIFF
--- a/hal/kubos-hal-iobc/src/lib.rs
+++ b/hal/kubos-hal-iobc/src/lib.rs
@@ -114,13 +114,28 @@ fn convert_raw_version(raw: ffi::supervisor_version) -> SupervisorVersion {
 
 /// Interface for retrieving iOBC supervisor version data
 pub fn supervisor_version() -> Result<SupervisorVersion, String> {
+    /*
     let mut version: ffi::supervisor_version = unsafe { mem::uninitialized() };
     let version_result = unsafe { ffi::supervisor_get_version(&mut version) };
     if !version_result {
         Err(String::from("Problem retrieving supervisor version"))
     } else {
         Ok(convert_raw_version(version))
-    }
+    }*/
+    Ok(SupervisorVersion {
+        dummy: 1,
+        spi_command_status: 2,
+        index_of_subsystem: 3,
+        major_version: 4,
+        minor_version: 5,
+        patch_version: 6,
+        git_head_version: 7,
+        serial_number: 8,
+        compile_information: vec![9; ffi::LENGTH_COMPILE_INFORMATION],
+        clock_speed: 10,
+        code_type: 11,
+        crc: 12
+    })
 }
 
 /// Converts raw bytes from iOBC into SupervisorHousekeeping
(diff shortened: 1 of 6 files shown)


hal/kubos-hal-iobc/src/lib.rs
```diff
@@ -114,13 +114,28 @@ fn convert_raw_version(raw: ffi::supervisor_version) -> SupervisorVersion {
...
 pub fn supervisor_version() -> Result<SupervisorVersion, String> {
+    /*
     let mut version: ffi::supervisor_version = unsafe { mem::uninitialized() };
     let version_result = unsafe { ffi::supervisor_get_version(&mut version) };
     if !version_result {
         Err(String::from("Problem retrieving supervisor version"))
     } else {
         Ok(convert_raw_version(version))
-    }
+    }*/
```
plauche: This is (horrible) temp code to expose supervisor version data.

plauche:
Current dilemmas - 

1. Regarding recreating the `SupervisorVersion` model in this program. This currently must be done because both the GraphQL traits and the `kubos_hal_iobc::SupervisorVersion` type are external to this crate. The Rust compiler rules [[1]](https://doc.rust-lang.org/error-index.html#E0117) [[2]](https://github.com/rust-lang/rfcs/blob/master/text/1023-rebalancing-coherence.md) dictate that external traits may not be applied to external types. The current solution is to create a new wrapper type, expose attributes of the original type through functions and apply the GraphQL traits to that new type. An alternative solution would be to apply the GraphQL traits directly to `kubos_hal_iobc::SupervisorVersion`, but I'm not particularly keen on bringing the GraphQL dependency down into the driver layer.

2. GraphQL numerical types are limited to 32-bit signed `Integer` and signed `Float`. This is not a limitation of Juniper but part of the GraphQL spec [[1]](http://facebook.github.io/graphql/October2016/#sec-Int) [[2]](http://facebook.github.io/graphql/October2016/#sec-Float). The majority of the data returned back from the supervisor is in the form of unsigned integers. The current solution it to convert data to 32-bit signed integers before handing off to GraphQL. Do we foresee any problems with this?

pengwynn:
> Regarding recreating the SupervisorVersion model in this program. This currently must be done because both the GraphQL traits and the kubos_hal_iobc::SupervisorVersion type are external to this crate. 

Not a Rust dev (yet?), but that seems sane given the constraints.

> The current solution it to convert data to 32-bit signed integers before handing off to GraphQL

If we're worried about data loss we could explore a [custom scalar](https://www.apollographql.com/docs/graphql-tools/scalars.html#examples) (which [Juniper seems to support](https://github.com/graphql-rust/juniper/blob/master/juniper/src/macros/scalar.rs)). While that might work for backend things, I'm guessing the value will ultimately make it to the UI and JavaScript can't handle integers of that size so we'd have to coerce to a String or other representation anyway.

plauche:
[iOBC Supervisor Handler](https://trello.com/c/wayX3uIo/61-iobc-supervisor-handler)

plauche:
Closed in favor of https://github.com/kubos/kubos/pull/24

PR #231, opened 2018-01-19
Adding blurbs about partuuid process and IDs

CatGarab:
PRs kubos/kubos-linux-build#83 and kubos/uboot#22 change the way that the BBB and MBM2 boot up. I was going to change the relevant documentation, but apparently that was never documented...

So instead I'm adding a new blurb about how U-Boot gets the boot device for Linux and then the new disk signatures that are used to generate the partuuids.


CatGarab:
[BBB - handle only eMMC](https://trello.com/c/QarGCPuH/147-bbb-handle-only-emmc)

CatGarab:
Closing pull request and reopening against new kubos/kubos repo

